PR #60, opened 2022-01-17
Allow specifying EF_PD_MCAST_LOOP via EF_VI_PD_FLAGS env var

vsolontsov-volant:
This PR is to let a user specifying more protection domain flags via `EF_VI_PD_FLAGS` environment variable.
Potentially it can add h/w loopback support into `solar_replay`.

### DIFF
--- a/src/lib/ciul/pd.c
+++ b/src/lib/ciul/pd.c
@@ -19,22 +19,47 @@
 
 #include <net/if.h>
 
+static int __ef_tok_eq(const char* tok, size_t tok_len, const char* tmpl)
+{
+    size_t l = strlen(tmpl);
+    return l == tok_len && ! strncmp(tok, tmpl, l);
+}
+
+static enum ef_pd_flags __ef_env2flags(enum ef_pd_flags flags)
+{
+    const char* s = getenv("EF_VI_PD_FLAGS");
+    if( s == NULL )
+        return flags;
+
+    enum ef_pd_flags new_flags = 0;
+    const char* tok_end;
+    do
+    {
+        tok_end = strchr(s, ',');
+        if( ! tok_end )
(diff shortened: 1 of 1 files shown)


ol-alexandra (approved):
The commit log should be copied from the request subject, but otherwise looks reasonable.

vsolontsov-volant:
> The commit log should be copied from the request subject, but otherwise looks reasonable.

Thanks a lot for your review. I fixed the comment, so should be fine.
Please let me know if there's anything I can do to push it through.

src/lib/ciul/pd.c
```diff
@@ -19,22 +19,47 @@
...
+        else if( __ef_tok_eq(s, tok_end - s, "phys") )
+            new_flags |= EF_PD_PHYS_MODE;
+        else if( __ef_tok_eq(s, tok_end - s, "default") )
+            flags = 0;
+        else if( __ef_tok_eq(s, tok_end - s, "mcast_loop") )
+            new_flags |= EF_PD_MCAST_LOOP;
+        s = tok_end + 1;
+    } while( *tok_end != '\0' );
+
+    return new_flags != 0 ? new_flags : flags;
```
rhughes-xilinx, line 52: I think this has changed the behaviour of `EF_VI_PD_FLAGS=default`

vsolontsov-volant: Not really (unless I'm missing something).
Stand-alone `default` resets the `flags` to zero (not `new_flags`).

rhughes-xilinx: Oh yes, so it does.

rhughes-xilinx (approved):

ol-alexandra:
I've pushed this change to the master branch, which will be released as onload-8.0.  Eventually.

PR #90, opened 2022-07-28
issue #89: fix onload module file being removed during upgrades

### LINKED ISSUE #89: Onload rpm scriptlet is removing /usr/local/modules-load.d/onload.conf after an upgrade
Hi,

There's an issue during the upgrade process of an onload  rpm (built using specfile `scripts/onload_misc/openonload.spec`)

The standard rpm upgrade sequence is to install the new package as a duplicate on the system. We now have the 2 (old & new) versions, and once install of the new is ok, remove the old one.
The issue lies in the rpm scriptlets: an upgrade of a package is seen as:
- Install the new version of the package (thus, triggers `%pre / %post` scripts)
- Remove the old version of the pacakge (thus, triggers `%preun / %postun` scripts)

This behavior can be tested with `rpm --reinstall -vv onload 2>&1| grep -ie modules-load.d`. The `cp` will happen before the `rm`.

As specified in [Packaging-Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_syntax) there should be a test wether it's a full-blown install/uninstall, or if it's a package upgrade.

Saruspete:
Fixes #89 

As specified in [Packaging-Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_syntax) there should be a test wether it's a full-blown install/uninstall, or if it's a package upgrade.

I only added it on the post-uninstall sequence, as we still want to have the latest version of `/usr/share/onload/onload_modules-load.d.conf` to be propagated on the copied (untracked) file.

### DIFF
--- a/scripts/onload_misc/openonload.spec
+++ b/scripts/onload_misc/openonload.spec
@@ -228,11 +228,13 @@ ldconfig -n /usr/lib /usr/lib64
 
 %postun
 
-if [ `cat /proc/1/comm` == systemd ]
-then
-  rm /usr/local/lib/modules-load.d/onload.conf
-else
-  rm /etc/sysconfig/modules/onload.modules
+# Remove these files only during uninstall, not during an upgrade
+if [ $1 == 0 ]; then
+  if [ `cat /proc/1/comm` == systemd ]; then
+    rm /usr/local/lib/modules-load.d/onload.conf
+  else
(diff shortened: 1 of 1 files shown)


tcrawley-xilinx (approved):
Looks good!
Thanks for contributing, I have tested this separately and it fixes the error described.

PR #92, opened 2022-08-16
implement capability to disable netdriver build

ol-sergeyn:
Introduce the HAVE_SFC variable to control sfc/driverlink build.
If it is unset (default), disable build of the entire netdriver
and driverlink support, and also other code in sfc_resource module
which uses driverlink API:
- PIO support
- EF100 slice extensions support

It is useful to isolate netdriver and onload modules build to get an easier
way to add support of new kernel versions in case when sfc build is broken
on new kernels (5.18 for now). For AF_XDP we do not need sfc and driverlink,
so we may proceed with new kernel support avoiding netdriver issues.

---

**Testing done**
Onload works with Debian 5.18.0-2-amd64 in AF_XDP mode when building with `MMAKE_LIBERAL=1`. (despite a lot of warnings when building modules).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,8 @@ Options:
   V=1              Print full build command lines
   NDEBUG=1         Create optimized build
   MMAKE_LIBERAL=1  Turn off -Werror
+  HAVE_SFC=0       Build without netdriver and driverlink support to work in
+                   AF_XDP mode only
   BUILD_PROFILE=x  Onload configuration, e.g. "cloud" (default: extra)
   KBUILDTOP=<path> Where to put driver build (default: build/$$KARCH_linux-$$KVER)
   KPATH=<path>     Kernel to build for (default: /lib/modules/`uname -r`/build)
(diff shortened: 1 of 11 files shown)


ol-alexandra (approved):

sianj-xilinx:
Thanks @ol-sergeyn . I think that the way we'd prefer to do this would be to invert the sense of the defines, so include the ef10/ef100 code based on something like `HAVE_SFC`. That makes it consistent with how we're handling the presence or absence of the X3 net driver.

ol-sergeyn:
> Thanks @ol-sergeyn . I think that the way we'd prefer to do this would be to invert the sense of the defines, so include the ef10/ef100 code based on something like `HAVE_SFC`. That makes it consistent with how we're handling the presence or absence of the X3 net driver.


Thank you for your reply.
I've updated the patch. Please check it out.

Now when you just call `make` it is built without sfc. To build with sfc one should call `HAVE_SFC=1 make`

ol-sergeyn:
I also added the `--no-sfc` option into `onload_install` script to have an ability to skip sfc modules installing if they were not built.

sianj-xilinx (commented):
Thanks @ol-sergeyn. This looks good, but to merge it I think we'd need to default to HAVE_SFC being 1 if it's not explicitly set to 0, to avoid breaking existing builds.

ol-sergeyn:
> Thanks @ol-sergeyn. This looks good, but to merge it I think we'd need to default to HAVE_SFC being 1 if it's not explicitly set to 0, to avoid breaking existing builds.

Fixed

PR #93, opened 2022-08-18
Fix x86 find_syscall_table() with new IBRS linux code

ol-alexandra:
New enhancements in Spectre mitigation in Linux kernel broke
find_syscall_table() function by adding a few bytes of asm instructions
in the middle of the code we are parsing.

Do not assume that "call do_syscall_64" goes right after saving
the syscall number, to match the code from linux>=5.18.14 and
Debian's backport in 5.10.0-17.

---
**Tested** with Debian `5.10.0-17-amd64` and Ubuntu `5.15.0-46-generic`.  Both fail without this patch with:
```
[ 2180.936513] [sfc efrm] find_syscall_table: didn't find do_syscall_64()
[ 2180.943123] [sfc efrm] init_sfc_resource: ERROR: failed to find syscall table
```
Both basically work with this patch.  Given that sapi-ts is now opensource I've checked that the sanity tests are green.

Tested with `retbleed=unret` & `retbleed=ibpb` kernel parameter (with Debian kernel).  These parameters make `UNTRAIN_RET` to be a call.

I guess that together with #92 this patch would fix Onload for linux>=5.18.14.  Without that Sergey's patch I'm unable to test linux-5.18 because the net driver fails to build (and I assume that proposing a fix for net driver via Onload pull request is not a way to get the patch merged).

### DIFF
--- a/src/driver/linux_resource/syscall_x86.c
+++ b/src/driver/linux_resource/syscall_x86.c
@@ -147,19 +147,54 @@ void** find_syscall_table(void)
    *    48 89 e7
    * movslq	%eax, %rsi
    *    48 63 f0
+   * linux>=5.18.14: "clobbers %rax" with IBRS in the middle
    * call	do_syscall_64
    *    e8 XX XX XX XX
    *
+   * linux>=5.18.14 and various backports (Debian's 5.10.0-17-amd64,
+   * Ubuntu's 5.15.0-46) add some code before the call instruction above.
+   * The code is: IBRS_ENTER + UNTRAIN_RET with a comment
+   * "clobbers %rax, make sure it is after saving the syscall nr"
+   *
+   * For now we just look for the first "e8" byte, but it is very
+   * error-prone.  Such a "e8" byte may be data, offset or something like
+   * this.  However it is hard to find a better solution, because these
+   * 2 macros depends on the kernel config option, and then the asm code is
+   * hot-patched at load time depending on CPU properties.
    */
   p += 0x40; /* skip the first part of entry_SYSCALL_64() */
   result = 0;
   pend = p + 1024 - 11;
   while (p < pend) {
-    if( p[0] == 0x48 && p[1] == 0x89 && p[3] == 0x48 && p[6] == 0xe8 &&
+#if 0
+    if( p[0] == 0x48 )
+      EFRM_ERR("%02x %02x %02x %02x %02x %02x %02x %02x %02x %02x %02x %02x "
+               "%02x %02x %02x %02x %02x %02x",
+               p[0], p[1], p[2], p[3], p[4], p[5], p[6], p[7], p[8], p[9],
+               p[10], p[11], p[12], p[13], p[14], p[15], p[16], p[17]);
+#endif
+    if( p[0] == 0x48 && p[1] == 0x89 && p[3] == 0x48 &&
         ((p[2] == 0xc7 && p[4] == 0x89 && p[5] == 0xe6) ||
          (p[2] == 0xe7 && p[4] == 0x63 && p[5] == 0xf0)) ) {
-      result = (unsigned long)p + 11;
-      result += p[7] | (p[8] << 8) | (p[9] << 16) | (p[10] << 24);
+      /* Skip IBRS: search for the nearest "call", 0xe8. */
+      unsigned char *p1 = p + 6;
+      while( *p1 != 0xe8 )
+        p1++;
+      /* Is this our "call do_syscall_64", or something from UNTRAIN_RET?
+       * In the second case these calls go one after another, and we should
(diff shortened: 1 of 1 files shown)


src/driver/linux_resource/syscall_x86.c
```diff
@@ -147,19 +147,41 @@ void** find_syscall_table(void)
...
+               "%02x %02x %02x %02x %02x %02x",
+               p[0], p[1], p[2], p[3], p[4], p[5], p[6], p[7], p[8], p[9],
+               p[10], p[11], p[12], p[13], p[14], p[15], p[16], p[17]);
+#endif
+    if( p[0] == 0x48 && p[1] == 0x89 && p[3] == 0x48 &&
         ((p[2] == 0xc7 && p[4] == 0x89 && p[5] == 0xe6) ||
          (p[2] == 0xe7 && p[4] == 0x63 && p[5] == 0xf0)) ) {
-      result = (unsigned long)p + 11;
-      result += p[7] | (p[8] << 8) | (p[9] << 16) | (p[10] << 24);
+      /* Skip IBRS: search for the nearest "call", 0xe8. */
```
rhughes-xilinx, line 179:
```
.macro UNTRAIN_RET
#if defined(CONFIG_CPU_UNRET_ENTRY) || defined(CONFIG_CPU_IBPB_ENTRY)
        ANNOTATE_UNRET_END
        ALTERNATIVE_2 "",                                               \
                      CALL_ZEN_UNTRAIN_RET, X86_FEATURE_UNRET,          \
                      "call entry_ibpb", X86_FEATURE_ENTRY_IBPB
#endif
.endm
```
Have you tested on a machine with IBPB? Our dellr650s (Ice Lake) report it

ol-alexandra: Yes, I have (`Intel(R) Xeon(R) CPU E3-1220 v6 @ 3.00GHz`, reports both `ibrs` & `ibpb`).  However I agree that it all should not work here.  I'll explore what's going on.

ol-alexandra: `ibpb` reported by `cpuinfo` is probably `X86_FEATURE_IBPB`.  `X86_FEATURE_ENTRY_IBPB` is a different beast:
```
retbleed_select_mitigation()
{
        ....
	case RETBLEED_MITIGATION_IBPB:
		setup_force_cpu_cap(X86_FEATURE_ENTRY_IBPB);
		mitigate_smt = true;
		break;
```

ol-alexandra: I have a feeling that I should use AMD to get this `call` from `UNTRAIN_RET`...

src/driver/linux_resource/syscall_x86.c
```diff
@@ -147,19 +147,49 @@ void** find_syscall_table(void)
...
+      unsigned char *p1 = p + 6;
+      while( *p1 != 0xe8 )
+        p1++;
+      /* Is this our "call do_syscall_64", or something from UNTRAIN_RET?
+       * In the second case these calls go one after another, and we should
+       * take the second one.
+       * No kernel I know about has a "call" after "call do_syscall_64",
+       * so 2 consecuent calls mean that do_syscall_64 is the second one. */
```
rhughes-xilinx, line 187: You're right that "consecuent" should be a word, but it isn't.

src/driver/linux_resource/syscall_x86.c
```diff
@@ -147,19 +147,49 @@ void** find_syscall_table(void)
...
          (p[2] == 0xe7 && p[4] == 0x63 && p[5] == 0xf0)) ) {
-      result = (unsigned long)p + 11;
-      result += p[7] | (p[8] << 8) | (p[9] << 16) | (p[10] << 24);
+      /* Skip IBRS: search for the nearest "call", 0xe8. */
+      unsigned char *p1 = p + 6;
+      while( *p1 != 0xe8 )
+        p1++;
+      /* Is this our "call do_syscall_64", or something from UNTRAIN_RET?
+       * In the second case these calls go one after another, and we should
+       * take the second one.
```
rhughes-xilinx, line 185: Can this comment please include the hardware configuration you found that produces the ENTRY_IBPB thing?

rhughes-xilinx (approved):
I've been wondering the feasibility of writing a function which can determine the length of an x86 instruction. For the limited case of instructions likely to be found in the kernel, I reckon it could be done in only 20-30 lines, and would solve your problem of false positives due to finding 0xe8 in an argument.

This'll do for now, though.

ol-alexandra:
Dear AMD developers, please merge this patch into the master branch.

rhughes-xilinx:
Pushed

PR #100, opened 2022-09-21
Update README.md with new kernel versions

ol-alexandra:
Add better separation between AF_XDP Onload and Native Onload. Document how to build and use Onload without SFC support.


ol-alexandra:
Interestingly, AMD developers approved and merged the code which adds new kernel support and new `HAVE_SFC` build switch.  But refuse to accept the documentation for these changes.  I'm puzzled.

ol-alexandra:
Thank you very much!

PR #112, opened 2022-11-10
Compat for Linux 6.0

ol-sergeyn:
Fix `struct iov_iter` processing in `f_ops->read_iter()` handler.
New 6.0 Linux changed the way it initializes `struct iov_iter` when passing it to read_iter().

**Testing**:
Onload build passes on CentOS 8: kernels `6.0.7-1.el8.elrepo.x86_64` and `4.18.0-408.el8.x86_64` both DEBUG and NDEBUG

Run socket tester test which calls read() from libc:
`--ool=socket_cache --ool=onload --tester-run=sockapi-ts/level5/fd_caching/fd_cache_nonblock_sync:env=VAR.env.peer2peer_two_iut,check_first=FALSE,nonblock_first=TRUE,nonblock_func=fcntl,func=read`
The test is green.
Before the patch it failed with ci_assert on 6.0:
```
[  629.748359] [onload] FAIL at /var/tmp/dnman/onload/src/include/ci/tools/iovec.h:34
[  629.756430] [onload] ci_assert(iovlen > 0)
[  629.756430] from /var/tmp/dnman/onload/src/include/ci/tools/iovec.h:34
```

### DIFF
--- a/src/driver/linux_resource/kernel_compat.sh
+++ b/src/driver/linux_resource/kernel_compat.sh
@@ -142,8 +142,6 @@ EFRM_HAS_FLUSH_DELAYED_FPUT	export	flush_delayed_fput	include/linux/file.h	fs/fi
 
 EFRM_IRQ_FREE_RETURNS_NAME	symtype	free_irq	include/linux/interrupt.h void *(unsigned int, void *)
 
-EFRM_HAS_ITER_TYPE	memtype	struct_iov_iter	iter_type	include/linux/uio.h	u8
-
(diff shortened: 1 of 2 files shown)


ol-alexandra (approved):

ivatet-amd (approved):
Thank you, @ol-sergeyn. I can see the first patch in this PR improves ee1706a by using the `iter_is_iovec()` predicate, and the second one handles single-segment iovec/`ITER_UBUF`. I built it against kernels 4.18, 5.15 and 6.0 (`HAVE_SFC=0`) to double-check.

Issue #114, opened 2022-11-20
Broken modules build on linux 6.1-rc: No rule to make target 'vmlinux.o', needed by 'modpost'


ol-sergeyn:
Modules build is broken on Linux 6.1-rc:
```
$ mmakebuildtree --driver                                                                                                            
Build tree made for linux as x86_64_linux-6.1.0-0-amd64  

$ make -C build/x86_64_linux-6.1.0-0-amd64/                                                                                          
make: Entering directory '/home/nikitins/work/v5/build/x86_64_linux-6.1.0-0-amd64'                                                                            
/home/nikitins/work/v5/src/driver/linux_resource/kernel_compat.sh -k /usr/src/linux-headers-6.1.0-0-amd64 -q > driver/linux_resource/autocompat.h             
make[1]: Entering directory '/home/nikitins/work/v5/build/x86_64_linux-6.1.0-0-amd64/lib'                                                                     
make[2]: Entering directory '/home/nikitins/work/v5/build/x86_64_linux-6.1.0-0-amd64/lib/citools'                                                             
make -C /usr/src/linux-headers-6.1.0-0-amd64 NDEBUG= GCOV= CC=cc  KBUILD_BUILTIN=1 KBUILD_EXTMOD=/home/nikitins/work/v5/build/x86_64_linux-6.1.0-0-amd64/lib/c
itools                                                                                                                                                        
make[3]: Entering directory '/usr/src/linux-headers-6.1.0-0-amd64'                                                                                            
  CC      /home/nikitins/work/v5/build/x86_64_linux-6.1.0-0-amd64/lib/citools/buddy.o                                                                        
...
  CC      /home/nikitins/work/v5/build/x86_64_linux-6.1.0-0-amd64/lib/citools/memleak_debug.o
  AR      /home/nikitins/work/v5/build/x86_64_linux-6.1.0-0-amd64/lib/citools/built-in.a
make[3]: *** No rule to make target 'vmlinux.o', needed by 'modpost'.  Stop.
make[3]: Leaving directory '/usr/src/linux-headers-6.1.0-0-amd64'
make[2]: *** [/home/nikitins/work/v5/src/lib/citools/mmake.mk:88: all] Error 2
make[2]: Leaving directory '/home/nikitins/work/v5/build/x86_64_linux-6.1.0-0-amd64/lib/citools'
make[1]: *** [/home/nikitins/work/v5/src/lib/mmake.mk:19: all] Error 2
[...]

Issue #115, opened 2022-11-21
AF_XDP: Out of VI instances with given attributes (-16)


ol-damirm:
I see the following error message, when trying to run application under Onload in AF_XDP mode:
```shell
onload nc -l 12345
oo:nc.openbsd[64020]: netif_tcp_helper_alloc_u: ERROR: Failed to allocate stack (rc=-16)
See kernel messages in dmesg or /var/log/syslog for more details of this failure

# Here are entries from /var/log/syslog
[41316.063462] [sfc efrm] efrm_vi_alloc: Out of VI instances with given attributes (-16)
[41316.074187] [sfc efrm] efrm_vi_alloc: Out of VI instances with given attributes (-16)
```
I haven't seen this before. Git bisect says it's a regression from a commit cf9a523901fa "ON-14118: Move accept_vi_constraints to efhw hal".

ol-alexandra:
The issue have been finally fixed, see https://github.com/Xilinx-CNS/onload/pull/119#issuecomment-1336503388

I do not see any button to close the issue on github.  Probably it is the thing to be done by AMD developers.

PR #117, opened 2022-11-23
Fix modules build on Linux 6.1 

ol-sergeyn:
The fix for #114 

---
Testing:
Build on Debian11 with `6.1.0-rc3` passes.
```
mmakebuildtree --driver && make -C build/x86_64_linux-6.1.0-rc3/ HAVE_SFC=0 
```

Run Socket Tester `usecases/send_recv` test. All iterations pass.

### DIFF
--- a/mk/platform/linux.mk
+++ b/mk/platform/linux.mk
@@ -7,7 +7,7 @@ endif
 
 LINUX		   := 1
 
-LINUX_VERSION_3	   := $(shell cat $(KPATH)/include/{linux/utsrelease.h,generated/utsrelease.h,linux/version.h} 2>/dev/null | sed 's/^\#define UTS_RELEASE \"\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/; t; d')
+LINUX_VERSION_3	   := $(shell cat $(KPATH)/include/generated/utsrelease.h 2>/dev/null | sed 's/^\#define UTS_RELEASE \"\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/; t; d')
 
 DRIVER		   := 1
 MMAKE_USE_KBUILD   := 1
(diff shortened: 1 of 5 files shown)


ol-alexandra (approved):

ol-alexandra:
Please note that these patches are applied to the master commit from Oct 26, 2022 7ae8f1a, because later AF_XDP is broken, see #115.

mk/platform/linux.mk
```diff
@@ -35,3 +39,11 @@ endif
 MMAKE_KBUILD_ARGS_CONST := -C $(KPATH) NDEBUG=$(NDEBUG) GCOV=$(GCOV) CC=$(CC)
 MMAKE_KBUILD_ARGS = $(MMAKE_KBUILD_ARGS_CONST) $(MMAKE_KBUILD_ARGS_DBG)
 
+# Check if $(1) is greater than or equal to $(2)
+int_ge = $(shell test $(1) -ge $(2) 2>/dev/null && echo 1)
```
ivatet-amd, line 43: What would be your opinion on using `sort` for the version comparison?
```
greater_version = $(shell printf "$(1)\n$(2)\n" | sort --version-sort | tail -1)

KBUILD_LIB_MAKE_TRG ?=
ifeq ($(call greater_version,$(LINUX_VERSION_3),6.0), $(LINUX_VERSION_3))
KBUILD_LIB_MAKE_TRG += $(lib_obj_path)
endif
```

It could eliminate `LINUX_VERSION_xyz` variables and hopefully contribute to readability?

ol-sergeyn: Looks nice, thank you.

ivatet-amd:
@ol-alexandra Thanks for referring to #115, but is that issue Linux 6.1-specific? If it is, I suppose we should get it fixed before merging this PR. Otherwise, I'm inclined to get it merged and schedule #115 afterwards.

ol-alexandra:
This pull request does not fix net driver.  Without a net driver fix, one can test these patches with AF_XDP Onload only.  And AF_XDP Onload is broken (on all kernels).  I.e. you can't test these patches on top of the master branch.

ivatet-amd:
OK, I got it, thanks. I was concerned about the build time issues this PR resolves for the latest kernel, while your concern is about the runtime, which is broken for AF_XDP regardless of the kernel. To validate the runtime for this PR, we need to address #115. Let me try to schedule it.

PR #123, opened 2022-12-15
Purge challenge_ack_limit, following linux-6.0

ol-alexandra:
In linux-6.0 challenge_ack_limit have been deprecated in 79e3602caa6f9d59c4f66a268407080496dae408 ,
following implementation of per-socket limitation of challenge ACK in f2b2c582e82429270d5818fbabe653f4359d7024 .

Onload already applies per-socket limitation for challenge ACK, see the very beginning of ci_tcp_send_challenge_ack().  So it is safe to follow linux and remove global/per-stack limitation for challenge ACKs.

Additionally, it removes following Onload warning on linux>=6.0:
config: ERROR - option challenge_ack_limit (2147483647) larger than maximum

----

I do not have any specific tests for this, but I've checked that it basically works.
And I do not see the error any more.

### DIFF
--- a/src/include/ci/internal/ip_shared_types.h
+++ b/src/include/ci/internal/ip_shared_types.h
@@ -739,12 +739,9 @@ typedef struct {
   ci_iptime_t tconst_pmtu_discover_recover;
 #define CI_PMTU_TCONST_DISCOVER_RECOVER (30*1000)
 
-  /* RFC 5961: limit for challenge ack packets per tick,
-   * derived from NI_OPTS(netif).challenge_ack_limit. */
-  ci_uint32 tconst_challenge_ack_limit;
-
   /* Rate limit for ACKs sent in response to invalid TCP packets,
-   * time period in ticks. */
+   * time period in ticks.
+   * It is used to limit challenge ACKs from RFC 5961 as well. */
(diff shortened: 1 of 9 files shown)


ivatet-amd (approved):
LGTM, but someone else should approve too, WRT to the retiring policies, e.g. removal of `EF_CHALLENGE_ACK_LIMIT` from the user guide, etc.

ol-alexandra:
Hi AMD developers, I hope you had merry Christmas, and the New Year will be happier than the previous.
Do you have any news about this patch?

krishd-amd (approved):

ivatet-amd:
Hi @ol-alexandra, sorry for the delay! I checked with the policies and would prefer to deprecate `EF_INVALID_ACK_RATELIMIT` gradually to stay on the safe side if somebody uses it (although I don't anticipate many users). I want to file another PR to handle `INT_MAX` from `challenge_ack_limit` gracefully, along with a deprecation warning. Hope it sounds good enough to you?

ol-alexandra:
Yes, it sounds good.  It would be nice to see your code as a public pull request if possible, please.

ivatet-amd:
@ol-alexandra Thanks can could you take a look at https://github.com/Xilinx-CNS/onload/pull/126?

PR #126, opened 2023-02-03
ON-14617: Deprecate EF_CHALLENGE_ACK_LIMIT

ivatet-amd:
The Linux kernel 6.0 since 79e3602caa6f disables the global challenge ACK rate limitation by default. It starts treating 2147483647 (i.e. `INT_MAX`) written to `/proc/sys/net/ipv4/tcp_challenge_ack_limit` as "unlimited".

Onload reads that value as the default for `EF_CHALLENGE_ACK_LIMIT`, and so to apply the same to Onload, this patch:

1. Extends the supported range of `EF_CHALLENGE_ACK_LIMIT` from [0..64K] to [0..`INT_MAX`]. Consequently, it addresses the following warning:
```
onload: config: ERROR - option challenge_ack_limit (2147483647) larger than maximum (65535)
```
2. Disables the rate limiting in `ci_tcp_send_challenge_ack()` when `EF_CHALLENGE_ACK_LIMIT` is `INT_MAX`, similarly to the Linux kernel.
3. Disables `EF_CHALLENGE_ACK_LIMIT` by default.
4. Adds a deprecation warning. We suggest using `EF_INVALID_ACK_RATELIMIT`, which exists in Linux kernel 3.19 since f2b2c582e824 as `/proc/sys/net/ipv4/tcp_challenge_ack_limit`.

A follow-up ChangeLog note is a separate patch. Additionally, Onload User Guide must reflect that.

Suggested-by: Alexandra Kossovsky <Alexandra.Kossovsky@oktetlabs.ru>

### DIFF
--- a/src/include/ci/internal/opts_netif_def.h
+++ b/src/include/ci/internal/opts_netif_def.h
@@ -860,7 +860,7 @@ CI_CFG_OPT("EF_CHALLENGE_ACK_LIMIT", challenge_ack_limit,
 "window attack mitigation, RFC 5961; in packets per second.  "
 "The limitation applies for each Onload stack separately.\n"
 "The value from /proc/sys/net/ipv4/tcp_challenge_ack_limit is used by default.",
-          , , CI_CFG_CHALLENGE_ACK_LIMIT, 0, 65535, count)
+          , , CI_CFG_CHALLENGE_ACK_LIMIT, 0, INT_MAX, count)
 
 CI_CFG_OPT("EF_INVALID_ACK_RATELIMIT", oow_ack_ratelimit, ci_uint32,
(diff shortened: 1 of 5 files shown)


ol-alexandra (approved):

ivatet-amd:
@ol-alexandra. I was trying to get away with as fewer `#include <limits.h>` as possible and could do 2 instead of 3 if included in opts_netif_def.h, but that file doesn't include anything, and I didn't want to spoil it. Redefining `INT_MAX` manually doesn't look appealing to me, either!

ol-alexandra:
Ideally `#include <limits.h>` should go to some platform or compat file, but sometimes it is not obvious how.

PR #127, opened 2023-02-03
Fix aarch64 sys_call_table look up code

ol-sergeyn:
Fix aarch64 `sys_call_table` look up code to be compatible with modern kernels.

Tested on Raspberry Pi 4 Model B Rev 1.4 with Debian 11 with kernel versions 5.15.83, 5.10.110, 5.4.83:
```bash
# Build and load is ok
sudo ./scripts/onload_install --no-sfc --no-initramfs
sudo onload_tool reload --onload-only

# Registering AF_XDP interface is ok
echo /home/pi/onload/src/tools/bpf_link_helper/bpf-link-helper | sudo tee /sys/module/sfc_resource/parameters/bpf_link_helper
echo eth0 | sudo tee /sys/module/sfc_resource/afxdp/register

# Running netcat (active/passive connections). Stacks are created successfully, data passes
echo 0 | sudo tee /sys/module/sfc_resource/parameters/enable_af_xdp_flow_filters
sudo onload nc -l -p 6666
sudo onload nc 10.0.0.2 6666
```

### DIFF
--- a/src/driver/linux_resource/syscall_aarch64.c
+++ b/src/driver/linux_resource/syscall_aarch64.c
@@ -10,11 +10,22 @@ void** efrm_syscall_table = NULL;
 EXPORT_SYMBOL(efrm_syscall_table);
 
 
-static typeof(aarch64_insn_decode_immediate) *aarch64_insn_decode_immediate_sym;
-static typeof(aarch64_insn_read) *aarch64_insn_read_sym;
-static typeof(aarch64_insn_extract_system_reg) *aarch64_insn_extract_system_reg_sym;
-static typeof(aarch64_get_branch_offset) *aarch64_get_branch_offset_sym;
-static typeof(aarch64_insn_adrp_get_offset) *aarch64_insn_adrp_get_offset_sym;
+/* The kernel contains some neat routines for doing AArch64 code inspection.
+ * Some of them are available as inline routines, but some are unfortunately
+ * not exported to modules.
+ * So we use efrm_find_ksym() to resolve these functions if it is available.
+ * Otherwise we copy their implementations here.
+ */
+#ifdef EFRM_HAVE_NEW_KALLSYMS
+static typeof(aarch64_insn_decode_immediate) *ci_aarch64_insn_decode_immediate;
+static typeof(aarch64_insn_read) *ci_aarch64_insn_read;
+static typeof(aarch64_insn_extract_system_reg) *ci_aarch64_insn_extract_system_reg;
+static typeof(aarch64_get_branch_offset) *ci_aarch64_get_branch_offset;
+static typeof(aarch64_insn_adrp_get_offset) *ci_aarch64_insn_adrp_get_offset;
+#else
+#include "ci_arm64_patching.h"
+#include "ci_arm64_insn.h"
+#endif /* EFRM_HAVE_NEW_KALLSYMS */
 
 /* Depending on the kernel version, locating the syscall table may be more
  * or less straigthforward.
@@ -62,8 +73,11 @@ find_el_sync(void)
 {
   ci_uint8 *vectors;
   ci_uint32 insn;
-  ci_uint8 *el_sync = efrm_find_ksym("el0_sync");
+  ci_uint8 *el_sync = NULL;
 
+#ifdef EFRM_HAVE_NEW_KALLSYMS
+  el_sync = efrm_find_ksym("el0_sync");
+#endif
   if (el_sync != NULL)
     return el_sync;
 
@@ -75,7 +89,7 @@ find_el_sync(void)
   vectors += 5 * sizeof(ci_uint32);
 #endif
 #endif
-  if (aarch64_insn_read_sym(vectors, &insn)) {
+  if (ci_aarch64_insn_read(vectors, &insn)) {
     EFRM_WARN("%s: cannot read vbar_el1", __func__);
     return NULL;
   }
@@ -83,7 +97,7 @@ find_el_sync(void)
     EFRM_WARN("%s: el0_sync entry is not a branch", __func__);
     return NULL;
   }
-  el_sync = vectors + aarch64_get_branch_offset_sym(insn);
+  el_sync = vectors + ci_aarch64_get_branch_offset(insn);
 
   return el_sync;
 }
@@ -91,10 +105,13 @@ find_el_sync(void)
 static ci_uint8 *
 find_el_svc_entry(void)
 {
-  ci_uint8 *el_svc = efrm_find_ksym("el0_svc");
+  ci_uint8 *el_svc = NULL;
   ci_uint8 *el_sync;
   ci_uint32 insn;
 
+#ifdef EFRM_HAVE_NEW_KALLSYMS
+  el_svc = efrm_find_ksym("el0_svc");
+#endif
   if (el_svc != NULL)
     return el_svc;
   el_sync = find_el_sync();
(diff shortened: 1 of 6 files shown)


ol-alexandra (approved):

ivatet-amd:
Shortlisting for the next week, sorry for the delay.

src/driver/linux_resource/syscall_aarch64.c
```diff
@@ -6,15 +6,191 @@
...
+#else
+/* X-SPDX-Source-URL: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git */
+/* X-SPDX-Source-Tag: v5.15.83 */
+/* X-SPDX-Source-File: arch/arm64/kernel/patching.c */
+/* X-SPDX-Source-File: arch/arm64/lib/insn.c */
+/* X-SPDX-License-Identifier: GPL-2.0-only */
+/* X-SPDX-Comment: Functions we need to inspect aarch64 instructions for the
+ *                 case when efrm_find_ksym() is unavailable. "ci" prefix is
+ *                 added to all the functions. */
+#include <asm/patching.h>
```
ivatet-amd, line 40: I think in that case we need to remove this pair of include's because the kernels that don't have the copied C files (patching.c and insn.c) also don't seem to have the corresponding H files. However, please check the above comment: maybe we want to have separate C and H files?

Build-time check (Linux 5.10.35 GCC 9 Ubuntu 22):
```
  CC [M]  /root/iteterev/onload_internal/build/aarch64_linux-5.10.35-2.0.0.0/src/driver/linux_resource/syscall_aarch64.o
/root/iteterev/onload_internal/src/driver/linux_resource/syscall_aarch64.c:40:10: fatal error: asm/patching.h: No such file or directory
   40 | #include <asm/patching.h>
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
```

src/driver/linux_resource/syscall_aarch64.c
```diff
@@ -6,15 +6,191 @@
...
+#ifdef EFRM_HAVE_NEW_KALLSYMS
+static typeof(aarch64_insn_decode_immediate) *ci_aarch64_insn_decode_immediate;
+static typeof(aarch64_insn_read) *ci_aarch64_insn_read;
+static typeof(aarch64_insn_extract_system_reg) *ci_aarch64_insn_extract_system_reg;
+static typeof(aarch64_get_branch_offset) *ci_aarch64_get_branch_offset;
+static typeof(aarch64_insn_adrp_get_offset) *ci_aarch64_insn_adrp_get_offset;
+#else
+/* X-SPDX-Source-URL: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git */
+/* X-SPDX-Source-Tag: v5.15.83 */
+/* X-SPDX-Source-File: arch/arm64/kernel/patching.c */
```
ivatet-amd, line 34: Have you considered splitting these into separate files, e.g. 2x .C and 2x .H? For instance, we could add them as-is in one commit and trim them down to what's needed for this PR in the subsequent commit. Perhaps, this could add some clarity?

Given that the patch prefixes the kernel functions with `ci_`, we aren't worrying about the name clashing in newer kernels where these files exist. I suppose if they had the same/unchanged name, we would have to handle three cases: (1) old kernels with `kallsyms_on_each_symbol()` and `EFRM_HAVE_NEW_KALLSYMS`, (2) mid-range, e.g. 5.7..5.13 without neither approach and (3) newer kernels with the patching.c and others. Is it why this PR prefixes them with `ci_`?

src/driver/linux_resource/syscall_aarch64.c
```diff
@@ -57,70 +233,211 @@ static typeof(aarch64_insn_adrp_get_offset) *aarch64_insn_adrp_get_offset_sym;
...
 static ci_uint8 *
 find_el_sync(void)
 {
   ci_uint8 *vectors;
   ci_uint32 insn;
-  ci_uint8 *el_sync = efrm_find_ksym("el0_sync");
+  ci_uint8 *el_sync = NULL;
 
+#ifdef EFRM_HAVE_NEW_KALLSYMS
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 13, 0)
```
ivatet-amd, line 255: Hmm, I thought `EFRM_HAVE_NEW_KALLSYMS` was removed in 5.7.

src/driver/linux_resource/syscall_aarch64.c
```diff
@@ -57,70 +233,211 @@ static typeof(aarch64_insn_adrp_get_offset) *aarch64_insn_adrp_get_offset_sym;
...
 }
 
+static ci_uint8 *
+find_next_bl(ci_uint8 *ptr)
+{
+  ci_uint32 insn;
+
+  while (1) {
+    CI_AARCH64_INSN_READ(ptr, insn);
+    if (aarch64_insn_is_bl(insn))
```
ivatet-amd, line 311: nit: (This is a matter of preference, really) Shall we put it inside `while` like `while (!aarch64_insn_is_bl(insn))` here and other occurrences?

src/driver/linux_resource/syscall_aarch64.c
```diff
@@ -142,53 +461,122 @@ find_syscall_table_via_vbar(void)
...
+  el_svc = find_next_bl(el_svc);
+  CI_AARCH64_INSN_READ(el_svc, insn);
+
+  /* Jumping to do_el0_svc(). */
+  do_el0_svc = el_svc + ci_aarch64_get_branch_offset(insn);
+  TRAMP_DEBUG("%s:%d: 'bl do_el0_svc' = %px (%x)", __FUNCTION__, __LINE__,
+              el_svc, insn);
+  TRAMP_DEBUG("%s:%d: do_el0_svc = %px", __FUNCTION__, __LINE__, do_el0_svc);
+
+  /* Next (and the last) we must find the function call where sys_call_table
```
ivatet-amd, line 504: This PR is a decent dive into low-level machinery, and I'm afraid I can't allocate more CPU cycles to review it. I trust that it works correctly!

Generally, what symptoms do you observe if something in the new code goes off the rails? Is it an instant kernel panic?

ivatet-amd (changes requested):

src/driver/linux_resource/syscall_aarch64.c
ol-sergeyn: No, kernel won't panic, just stop module loading.

Example:
```bash
$ sudo insmod ./aarch64_linux-5.10.110-v8-ol+/driver/linux/sfc_resource.ko 
insmod: ERROR: could not insert module ./aarch64_linux-5.10.110-v8-ol+/driver/linux/sfc_resource.ko: Unknown symbol in module
```
dmesg:
```
[190409.448360] [sfc efrm] init_sfc_resource: RESOURCE driver starting
[190409.454665] [sfc efrm] find_el_sync: el0_sync entry is not a branch
[190409.461037] [sfc efrm] init_sfc_resource: ERROR: failed to find syscall table
```

src/driver/linux_resource/syscall_aarch64.c
ol-sergeyn: Thank you for pointing it out. Indeed, this extra code is not needed.

src/driver/linux_resource/syscall_aarch64.c
ol-sergeyn: Besides this compile error there are also runtime problems with linux 5.10, which need fixing.

ol-sergeyn: Removed those includes - we really do not need them.
Placed linux functions into separate c/h files.
Fixed the code to be compatible with linux-5.10.

src/driver/linux_resource/syscall_aarch64.c
ol-sergeyn: Fixed

src/driver/linux_resource/syscall_aarch64.c
ivatet-amd: Thanks, it's good we don't make the host unusable after a problematic run.

ivatet-amd:
Thanks for the latest PR - I'll have another look tomorrow and merge it. 👍

ivatet-amd (approved):

Issue #131, opened 2023-03-09
6.1.8-100.fc36: kernel BUG at mm/hugetlb.c:1123


ol-sergeyn:
Onload causes kernel BUG with call trace on Fedora with Linux 6.1:
```
[  391.824180] vma ffff9cf544af3af0 start 0000000000000000 end 0000000000000000 mm 0000000000000000                                                                                                                
[  391.824180] prot 0 anon_vma 0000000000000000 vm_ops 0000000000000000                                                                                                                                            
[  391.824180] pgoff 0 file ffff8a794a9f8e00 private_data 0000000000000000                                                                                                                                         
[  391.824180] flags: 0x0()                         
[  391.851261] ------------[ cut here ]------------                                                      
[  391.856419] kernel BUG at mm/hugetlb.c:1123!                                                          
[  391.861198] invalid opcode: 0000 [#1] PREEMPT SMP PTI                                                 
[  391.866834] CPU: 2 PID: 40582 Comm: nc Tainted: G           OE      6.1.8-100.fc36.x86_64 #1                                                                                                                    
[  391.876253] Hardware name: Dell Inc. PowerEdge R230/0FRVY0, BIOS 2.3.2 11/16/2017                                                                                                                               
[  391.884605] RIP: 0010:hugetlb_vm_op_close+0x174/0x1b0                                                 
[  391.890245] Code: 8b 80 d8 00 00 00 48 8b 00 4c 8b b0 30 02 00 00 e9 f8 fe ff ff 48 f7 de 48 89 df e8 c6 fc ff ff eb b24                                                                                        
[  391.911201] RSP: 0018:ffff9cf544af3a98 EFLAGS: 00010296                                               
[  391.917031] RAX: 00000000000000d6 RBX: ffffffff90cdf7e8 RCX: 0000000000000000                                                                                                                                   
[  391.924995] RDX: 0000000000000001 RSI: ffffffff8f74befb RDI: 00000000ffffffff                                                                                                                                   
[  391.932957] RBP: 0000000000000000 R08: 0000000000000000 R09: ffff9cf544af3908                                                                                                                                   
[  391.940919] R10: 0000000000000003 R11: ffffffff901465a8 R12: ffff9cf544af3af0                                                                                                                                   
[  391.948882] R13: 0000000000000002 R14: ffff8a79acfb0000 R15: ffff8a79acfb0d30                                                                                                                                   
[...]

Issue #129, opened 2023-02-24
Linux-6.1: HugeTLB: vma_lock already exists in hugetlb_vm_op_open


ol-sergeyn:
Linux prints this warning when creating an onload stack (non-AF_XDP):
`HugeTLB: vma_lock already exists in hugetlb_vm_op_open.`

How to reproduce:
- system: debian-testing with `6.1.0-2-amd64` kernel
- build and load Onload `mmakebuildtree && make && load.sh onload`
- run `onload nc -l 6666`

NB: sfc netdriver build is broken due to #124, so you need a patch to successfully build it and reproduce the issue:
```diff
diff --git a/src/driver/linux_net/drivers/net/ethernet/sfc/kernel_compat.h b/src/driver/linux_net/drivers/net/ethernet/sfc/kernel_compat.h
index a58c7f2e7f..e55a3d17a5 100644
--- a/src/driver/linux_net/drivers/net/ethernet/sfc/kernel_compat.h
+++ b/src/driver/linux_net/drivers/net/ethernet/sfc/kernel_compat.h
@@ -2681,4 +2681,9 @@ typedef atomic_t refcount_t;
 #define refcount_dec_and_test	atomic_dec_and_test
 #endif
 
+#ifdef EFX_HAVE_NAPI_ADD_WEIGHT
+/* linux >= 6.1 */
+#define netif_napi_add netif_napi_add_weight
+#endif
+
 #endif /* EFX_KERNEL_COMPAT_H */
diff --git a/src/driver/linux_net/drivers/net/ethernet/sfc/kernel_compat.sh b/src/driver/linux_net/drivers/net/ethernet/sfc/kernel_compat.sh
index 2e0eb7472d..49b2ba6e71 100755
--- a/src/driver/linux_net/drivers/net/ethernet/sfc/kernel_compat.sh
+++ b/src/driver/linux_net/drivers/net/ethernet/sfc/kernel_compat.sh
@@ -180,6 +180,7 @@ EFX_HAVE_SKB_FRAG_TRUESIZE		symtype	skb_add_rx_frag		include/linux/skbuff.h	void
 EFX_HAVE_INNER_IP_HDR			symbol	inner_ip_hdr		include/linux/ip.h
 EFX_HAVE_INNER_TCP_HDR			symbol	inner_tcp_hdr		include/linux/tcp.h
 EFX_HAVE_INDIRECT_CALL_WRAPPERS		file				include/linux/indirect_call_wrapper.h
[...]

PR #138, opened 2023-04-14
Refactor hugepage allocation in the kernel

ivatet-amd:
This series introduces the new hugetlb library to allocate mappable 2MiB hugepages in the kernel. The key features:

* Better Linux 6.1 support: the new hugetlb library uses the EFCT approach that doesn't require VMA workarounds broken in 6.1.
* Code deduplication: Onload used to have packet buffers and EFCT doing the same thing... differently.
* Use of `pin_user_pages()` instead of `find_get_page()` as suggested in https://lkml.org/lkml/2023/4/11/1286.
* New IOCTL to map packet buffers.

Please see more details in the commit messages.

### DIFF
--- a/src/include/kernel_utils/hugetlb.h
+++ b/src/include/kernel_utils/hugetlb.h
@@ -1,0 +1,139 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/* X-SPDX-Copyright-Text: (c) Copyright 2023 Advanced Micro Devices, Inc. */
+/*
+ ****************************************************************************
+ *
+ * A library to allocate hugepages in the kernelspace
+ * (which is not encouraged and not straightforward).
+ *
+ * How it works: the userspace donates a memfd file to the kernel module via
+ * ioctl(). Then the library calls ftruncate(), fallocate(), mmap(),
+ * pin_user_pages() and munmap() to get an instance of a 2 MiB hugepage,
+ * which we can give to NIC after dma_map_single(). For older kernels without
+ * the memfd support, the library calls hugetlb_file_setup() directly.
+ *
+ * Ideally, the userspace should do the allocation itself, i.e. ftruncate(),
+ * fallocate(), mmap(), and then make an ioctl() with the virtual address.
+ * In this case, we become aligned with the kernel API, and the library
+ * degenerates into a single pin_user_pages() call. This may require
+ * significant workflow refactoring in Onload.
+ *
+ ****************************************************************************
+ */
+
+#ifndef __OO_HUGETLB_H__
+#define __OO_HUGETLB_H__
+
+#include <onload/common.h>
+#include <onload/atomics.h>
+
+#define OO_HUGEPAGE_SIZE (2 * 1024 * 1024)
+
+struct oo_hugetlb_allocator {
+	struct file *filp;
+	off_t offset;
+	atomic_t refcnt;
+};
+
+struct oo_hugetlb_page {
+	struct file *filp;
+	struct page *page;
+};
+
+/* Create/destroy the memory allocator. */
+
+/*
+ * oo_hugetlb_allocator_create - Create a hugepage allocator.
+ *
+ * Parameters:
+ *   fd:             A donated memfd file descriptor to use for hugepage
+ *                   allocation or -1 on systems without memfd_create().
+ *
+ * Return:
+ *   On success, get a file reference identified by the file descriptor
+ *   or open a pseudo file with hugetlb_file_setup(), and return a valid
+ *   pointer to use later for allocation.
+ *
+ *   On failure, return a negative error number:
+ *     ENOSYS: Unable to find or call hugetlb_file_setup() in absence of memfd.
+ *     ENOMEM: Kernel memory allocation failure.
+ *     EINVAL: User error, e.g. a wrong file descriptor.
+ *     Those, returned by hugetlb_file_setup().
+ *
+ * Notes:
+ *   EINVAL should be treated as a fatal error indicating a software defect.
+ */
+extern struct oo_hugetlb_allocator *oo_hugetlb_allocator_create(int fd);
+
+extern struct oo_hugetlb_allocator *
+oo_hugetlb_allocator_get(struct oo_hugetlb_allocator *);
+
+extern void oo_hugetlb_allocator_put(struct oo_hugetlb_allocator *);
+
+static inline void
+oo_hugetlb_page_reset(struct oo_hugetlb_page *page)
+{
+	page->filp = NULL;
+	page->page = NULL;
+}
+
+/* Allocate/free one hugepage. */
+
+extern int
+oo_hugetlb_page_alloc_raw(struct oo_hugetlb_allocator *,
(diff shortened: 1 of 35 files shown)


ivatet-amd:
Hello @ol-alexandra and @ol-sergeyn,

I feel like sharing this pull request with you guys because Alexandra has implemented lots of related code, and Sergey reported a regression recently.

It is quite a large diff, but fundamentally I aim to address https://github.com/Xilinx-CNS/onload/issues/129 and https://github.com/Xilinx-CNS/onload/issues/131 by replacing the `shmget`/`shmat` approach with `memfd`, and to deduplicate the code. The rest of the pull request is plumbing. In particular, I have found propagating the allocator through `tcp_helper_resource_t` and `tcp_helper_cluster_t` non-trivial.

In general, I am happy to merge as long as you do not have objections against the implementation in `src/lib/kernel_utils/hugetlb.c`. Hope you find it entertaining. 👍

ol-alexandra:
I have a question.  Does this new approach help AF_XDP?  I.e. currently Onload-on-AF_XDP disables huge page support.  Can it change with the new approach?

I'll look at your patches, thank you for sharing them.

ivatet-amd:
> Does this new approach help AF_XDP?

If the initial problem with the hugepages and AF_XDP was VMA operations and page fault handling, I anticipate it is possible to apply the new hugetlb library to AF_XDP (with API additions/changes to handle more than one page at a time), though I did not do it in the PR. I have just tried to prototype it https://github.com/Xilinx-CNS/onload/commit/e3765e74f7c85b9ff0f7e8af0fd7b937f14c33b3 but ran out of time quota for today!

ivatet-amd:
> I have just tried to prototype it https://github.com/Xilinx-CNS/onload/commit/e3765e74f7c85b9ff0f7e8af0fd7b937f14c33b3 but ran out of time quota for today!

Have also tried to use `hugetlb_pages_prealloc()` in `xdp_register_umem()`, but the control flow still fails somewhere shortly afterwards. I can't continue with the AF_XDP for now, but still can't see reasons why hugetlb wouldn't work for AF_XDP.

src/include/kernel_utils/hugetlb.h
```diff
@@ -0,0 +1,139 @@
...
+
+#define HUGEPAGE_SIZE (2 * 1024 * 1024)
+
+struct hugetlb_allocator {
+	struct file *filp;
+	off_t offset;
+	atomic_t refcnt;
+};
+
+struct hugetlb_page {
```
ol-alexandra, line 38: I dislike the names.  10 years later, Linux kernel will introduce `hugetlb_page` structure, and we'll get a new compat issue out of nowhere.  Please add `oo_` prefix to all your names, or something like this.

ol-alexandra (changes requested):
Do I understand it correctly that your `Add kernel hugetlb library` patch b38a9443fa625f57 just copies EFCT code, reworking it to create a new clear API?  If so, could you state it clearly in the commit log?

Generally, I think your patches are sensible.

src/lib/efthrm/tcp_helper_cluster.c
```diff
@@ -1504,6 +1519,7 @@ int efab_tcp_helper_reuseport_bind(ci_private_t *priv, void *arg)
       drop_flags = 0;
     rc = thc_alloc_thr(thc, trb->cluster_restart_opt,
                        trb->cluster_hot_restart_opt,
+                       /* pktbuf_memfd */ -1,
                        &ni->opts,
                        ni->flags | drop_flags,
                        &thr);
```
ol-alexandra, line 1525: I read it as "Let's break nginx profile", because nginx works via this part of code and specifies `EF_USE_HUGE_PAGES` to 2.

src/lib/transport/ip/netif_init.c
```diff
@@ -2393,17 +2400,16 @@ static void init_resource_alloc(ci_resource_onload_alloc_t* ra,
...
   ra->in_memfd = -1;
-#ifdef MFD_HUGETLB
   /* The kernel code can cope with no memfd being provided, but only on older
    * kernels */
   {
     char mfd_name[CI_CFG_STACK_NAME_LEN + 8];
     snprintf(mfd_name, sizeof(mfd_name), "efct/%s", name);
-    ra->in_memfd = memfd_create(mfd_name, MFD_CLOEXEC | MFD_HUGETLB);
+    ra->in_memfd = syscall(__NR_memfd_create, mfd_name,
+                           MFD_CLOEXEC | MFD_HUGETLB);
```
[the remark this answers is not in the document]
ivatet-amd: Yup, added. I've also moved the memfd definitions in `ci/tools/platform/unix.h` to deduplicate, however, please let me know if there's a better place for these.

src/include/kernel_utils/hugetlb.h
ivatet-amd: I'm on the same page here. I didn't like the "crude" `hugetlb_xyz` names but couldn't think of a good prefix. Please find these prefixed with `oo_` in the latest PR incarnation.

ol-alexandra (approved):
LGTM, if you are happy with nginx.

ivatet-amd:
> Do I understand it correctly that your `Add kernel hugetlb library` patch [b38a944](https://github.com/Xilinx-CNS/onload/commit/b38a9443fa625f57bf30fd3d4097e396b07e4b0c) just copies EFCT code, reworking it to create a new clear API? If so, could you state it clearly in the commit log?

Yes, more or less, modulo replacing `find_get_pages()` with `pin_user_pages()`. I've tried to make it clearer in the commit message in the latest git push.

src/lib/efthrm/tcp_helper_cluster.c
ivatet-amd: Thanks for pointing that out! It does look like a regression in Linux 5.7+. I'm looking at how to donate a memfd file descriptor to this workflow. I didn't do it in one batch with other suggestions, including internal, to reduce the number of changes in flight. Thanks again.

ivatet-amd: Actually, it's probably a false alarm.

I remember fixing it specifically for `nginx` or `wrk` by introducing the allocator's reference counting for this and other clustering workflows:
```
-> efab_tcp_helper_reuseport_bind()
-> thc_alloc_thr()
-> tcp_helper_rm_alloc()
```
Where we bump a reference if the allocator is available:
```
  if( thc && thc->thc_pktbuf_alloc ) {
    rs->thc_pktbuf_alloc = oo_hugetlb_allocator_get(thc->thc_pktbuf_alloc);
```
The main problem was the undefined order of who is created first (`tcp_helper_cluster_t` or `tcp_helper_resource_t`) and then making sure that all `tcp_helper_resource_t` instances share the same allocator since they're the same cluster.

If there was a missing memfd donation, I would fix it by plumbing through `in_pktbuf_memfd`.

Let me add a comment in the code clarifying that: it's not obvious and I've already forgotten things enough so they look like a regression.

ivatet-amd:
Thanks for the prompt review! It's now merged.

PR #143, opened 2023-04-26
AMD + linux-6.2: find the syscall table

ol-alexandra:
Now Onload works on AMD CPU with linux-6.2.

### DIFF
--- a/src/driver/linux_resource/syscall_x86.c
+++ b/src/driver/linux_resource/syscall_x86.c
@@ -103,6 +103,81 @@ static bool is_movq_to_rdi(const unsigned char *p)
 }
 
 
+static unsigned char*
+find_syscall_table_from_do_syscall_64(unsigned char *my_do_syscall_64)
+{
+  unsigned char *p = my_do_syscall_64;
+  unsigned char *pend;
+  unsigned long result;
+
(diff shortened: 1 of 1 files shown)


src/driver/linux_resource/syscall_x86.c
```diff
@@ -253,39 +253,24 @@ void** find_syscall_table(void)
...
+        result = (unsigned long)p1 + 5;
+        result += p1[1] | (p1[2] << 8) | (p1[3] << 16) | (p1[4] << 24);
 
-      result = (unsigned long)p1 + 5;
-      result += p1[1] | (p1[2] << 8) | (p1[3] << 16) | (p1[4] << 24);
-      break;
+        ret = find_syscall_table_from_do_syscall_64((void*)result);
+        if( ret != NULL )
+          return (void**)ret;
+        p1++;
```
ivatet-amd, line 266: Not sure, but shall we skip five bytes instead of one if the address contains `0xe8` so that we don't call the `find_syscall_table_from_do_syscall_64()` predicate on data which is certainly not an address? Still, one byte doesn't look like a problem to me, but five seems a reasonable optimisation. What do you reckon?

ivatet-amd (approved):

PR #159, opened 2023-07-03
Handle XDP rings for linux>=6.3

ol-alexandra:
AF_XDP rings were allocated as physically-continuous memory before linux-6.3. Now rings are vmalloc'ed.  Let's map these new rings via pin_user_pages().

**Testing:**
Socket Tester sanity tests do not show any regression when using AF_XDP Onload with Debian Testing, kernels `6.3.0-0-amd64` and `6.1.0-9-amd64`

### DIFF
--- a/src/lib/efhw/af_xdp.c
+++ b/src/lib/efhw/af_xdp.c
@@ -151,6 +151,17 @@ struct event_waiter
   int budget;
 };
 
+/* Ring memory mapping artifacts */
+struct ring_map {
+  int n_pages;
+  struct page** pages;
+  void* vmapped_addr;
+};
+
+/* Are rings allocated as physically-continuous memory (linux<6.3) or
+ * via vmalloc (linux>=6.3)? */
+static bool rings_are_physically_continuous = true;
+
 /* Per-VI AF_XDP resources */
 struct efhw_af_xdp_vi
 {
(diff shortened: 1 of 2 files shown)


src/lib/efhw/af_xdp.c
```diff
@@ -151,6 +151,17 @@ struct event_waiter
...
+/* Ring memory mapping artifacts */
+struct ring_map {
+  int n_pages;
+  struct page** pages;
+  void* vmapped_addr;
+};
+
+/* Are rings allocated as physically-continuous memory (linux<6.3) or
+ * via vmalloc (linux>=6.3)? */
+static bool rings_are_physically_continuous = true;
```
mseymour-xilinx, line 163: I'm not sure we need this variable. I think it would be simpler to call `follow_pfn` for each ring unconditionally, and see whether it succeeds.

mseymour-xilinx (approved):

src/lib/efhw/af_xdp.c
ol-alexandra: I disagree with you.

These 2 cases are completely different.  It is lucky that (for now) we can determine the case by `follow_pfn`.  I'm not sure this hack will work forever.  And yes, I believe it is a hack.  So, I believe we need some sentinel which case is running here.  I'd prefer compile-time condition, but it is not easy.  Such a run-time variable is better than nothing.

ol-alexandra: Yes, that's my idea.

PR #166, opened 2023-08-15
linux_resource/kernel_compat_funcs: Fix hang when $KBUILD_MODULE_SYMVERS is empty

ZachG-gnu:
kernel_compat_funcs performs an awk on $KBUILD_MODULE_SYMVERS even when it is empty. This causes the script to hang forever. Add a check for if $KBUILD_MODULE_SYMVERS is empty before the awk.

### DIFF
--- a/src/driver/linux_resource/kernel_compat_funcs.sh
+++ b/src/driver/linux_resource/kernel_compat_funcs.sh
@@ -329,7 +329,7 @@ function test_export()
     #        May give a false positive if the export is conditional.
     #     3. The MAP file if present. May give a false positive
     #        because it lists all extern (not only exported) symbols.
-    if [ -f $KBUILD_MODULE_SYMVERS ]; then
+    if [[ $KBUILD_MODULE_SYMVERS && -f $KBUILD_MODULE_SYMVERS ]]; then
         if [ $efx_verbose = true ]; then
             echo >&2 "Looking for export of $symbol in $KBUILD_MODULE_SYMVERS"
 	fi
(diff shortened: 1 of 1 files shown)


abower-amd (approved):

src/driver/linux_resource/kernel_compat_funcs.sh
```diff
@@ -329,7 +329,7 @@ function test_export()
     #        May give a false positive if the export is conditional.
     #     3. The MAP file if present. May give a false positive
     #        because it lists all extern (not only exported) symbols.
-    if [ -f $KBUILD_MODULE_SYMVERS ]; then
+    if [[ $KBUILD_MODULE_SYMVERS && -f $KBUILD_MODULE_SYMVERS ]]; then
```
abower-amd, line 332: Sorry I was concentrating on whether bashisms were acceptable when I accepted this review but I think it's a bit messy duplicating the variable _and_ relying on not explicit behaviour, as this isn't specifically checking for being set.
I would prefer just putting the expansion in `"` in this case.

abower-amd (changes requested):

src/driver/linux_resource/kernel_compat_funcs.sh
abower-amd, line 332:
```suggestion
    if [ -f "$KBUILD_MODULE_SYMVERS" ]; then
```

abower-amd (approved):
Thanks for the contribution @ZachG-gnu!

PR #167, opened 2023-08-21
AF_XDP support for rx timestamps

wdebruij:
This is an initial demonstrator of how to support hardware receive timestamps in the af_xdp ef_vi.

Create a new XDP program that uses the recently added XDP rx metadata to (1) read the hardware timestamp if supported by the device and (2) store this in the AF_XDP buffer in headroom before the packet.

Update Onload to read this headroom and convert to the internal pkt timestamp fields.

To make this possible, also (1) add support for using externally compiled XDP programs and (2) add an optional map that advertises optional features of the XDP program.

### DIFF
--- a/src/tools/bpf_link_helper/xdp_onload_prepare.c
+++ b/src/tools/bpf_link_helper/xdp_onload_prepare.c
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: GPL-2.0
+
+/* xdp_onload_prepare -- load an XDP ELF object with xsk map and program
+ *
+ * cd $SOMEDIR
+ * git clone --recurse-submodules https://github.com/libbpf/bpftool.git
+ * cd bpftool
+ * export BPFTOOL=$PWD
+ * make
+ *
+ * cd $ONLOAD
+ * cd src/tools/bpf_link_helper
+ * gcc -Wall -o xdp_onload_prepare xdp_onload_prepare.c -L $BPFTOOL/libbpf/src -lbpf -I $BPFTOOL/libbpf/src/
+ */
+
+#include <errno.h>
+#include <error.h>
+#include <linux/if_link.h>
+#include <net/if.h>
+#include <stdlib.h>
+
+#include "libbpf.h"
+#include "bpf.h"
+
+int main(int argc, char **argv)
+{
+	struct bpf_object *obj;
+	struct bpf_program *prog;
+	struct bpf_map *map_xsk;
+	int prog_fd, ifindex;
+
+	if (argc != 3)
+		error(1, 0, "Usage: %s [ifname] [bpf_file]\n", argv[0]);
+
+	ifindex = if_nametoindex(argv[1]);
+	if (!ifindex)
+		error(1, errno, "if_nametoindex %s", argv[1]);
+
+	obj = bpf_object__open(argv[2]);
+	if (!obj)
+		error(1, errno, "bpf_object_open %s", argv[2]);
+
+	map_xsk = bpf_object__find_map_by_name(obj, "onload_xdp_xsk");
+	if (!map_xsk)
+		error(1, 0, "bpf_object__find_map_by_name onload_xdp_xsk");
+
+	prog = bpf_object__find_program_by_name(obj, "xdp_onload_prog");
+	if (!prog)
+		error(1, 0, "bpf_object__find_program_by_name");
+
+	bpf_program__set_ifindex(prog, ifindex);
+	bpf_program__set_flags(prog, BPF_F_XDP_DEV_BOUND_ONLY);
+
+	if (bpf_object__load(obj) < 0)
+		error(1, errno, "bpf_object__load");
+
+	prog_fd = bpf_program__fd(prog);
+	if (prog_fd < 0)
+		error(1, 0, "bpf_program__fd");
+
+	if (bpf_xdp_attach(ifindex, prog_fd, XDP_FLAGS_DRV_MODE, NULL))
+		error(1, 0, "bpf_program__attach_xdp");
+
+	if (bpf_object__pin_maps(obj, "/sys/fs/bpf"))
+		error(1, 0, "bpf_object__pin_maps");
+
+	printf("OK\n");
+	return 0;
+}
--- a/src/tools/bpf_link_helper/xdp_tstamp.c
+++ b/src/tools/bpf_link_helper/xdp_tstamp.c
@@ -1,0 +1,89 @@
+// SPDX-License-Identifier: GPL-2.0
+
+/* xdp_tstamp.c -- xdp program that reads optional hw rx timestamp
+ *
+ * See also https://docs.kernel.org/networking/xdp-rx-metadata.html
+ *
+ *
+ * # BUILD
+ *
+ * # install bpftool (if not installed yet)
+ * cd $SOMEDIR
+ * git clone --recurse-submodules https://github.com/libbpf/bpftool.git
+ * cd bpftool
+ * export BPFTOOL=$PWD
+ * make
+ *
+ * # install xdp-tools
+ * cd $SOMEDIR
+ * git clone https://github.com/xdp-project/xdp-tools.git
+ * cd xdp-tools
+ * XDPTOOLS=$PWD
+ * ./configure
+ * make
+ *
+ * cd $ONLOAD
+ * cd src/tools/bpf_link_helper
+ * clang -g -target bpf -O2 -I $BPFTOOL/src/libbpf/include -I XDPTOOLS/headers -c xdp_tstamp.c -o ./xdp_tstamp.o
+ *
+ *
+ * # RUN
+ *
+ * # insert the xdp prog *before* loading the onload kernel modules
+ * ./xdp_onload_prepare $DEV xdp_tstamp.o
+ */
+
+#include <linux/bpf.h>
+#include <bpf/bpf_helpers.h>
+#include <xdp/xdp_helpers.h>
+
+/* must match map created by onload/src/lib/efhw/af_xdp.c
+ *
+ * max_entries matches the number passed to
+ * /sys/module/sfc_resource/afxdp/register
+ */
+struct {
+	__uint(type, BPF_MAP_TYPE_XSKMAP);
+	__uint(key_size, sizeof(int));
+	__uint(value_size, sizeof(int));
+	__uint(max_entries, 4);
+} onload_xdp_xsk SEC(".maps");
+
+struct onload_xdp_rx_meta {
+#define ONLOAD_XDP_RX_META_TSTAMP 0x1
+	__u64 flags;
+	__u64 tstamp;
+};
+
+extern int bpf_xdp_metadata_rx_timestamp(const struct xdp_md *ctx,
+					 __u64 *timestamp) __ksym;
+
+SEC("xdp")
+int xdp_onload_prog(struct xdp_md *ctx)
+{
+	struct onload_xdp_rx_meta *meta;
+	void *data, *data_meta;
+	int ret;
+
+	ret = bpf_xdp_adjust_meta(ctx, -(int)(sizeof(struct onload_xdp_rx_meta)));
+	if (ret != 0)
+		return XDP_PASS;
+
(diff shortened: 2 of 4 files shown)


sianj-xilinx:
Firstly, thanks for this proposal. I'm about to disappear on holiday, but thought I'd stick a few initial thoughts up.

I was wondering whether there might be a better (but not yet existing) way to handle the configuration of timestamps that would fit in better with the existing onload mechanisms. I'm imagining something like being able to explicitly request timestamping on the AF_XDP socket using existing timestamping socket options. If we could determine at the time we set up the socket whether timestamping was supported this could then use the existing onload environment variables to enable timestamping, which include the ability to fail if the HW does not support timestamps. In general I'm keen on keeping required configuration by the user as minimal as possible, and having to provide an additional bpf map is a lot more complex than what we usually expect users to have to deal with. As far as I can tell there's not support in AF_XDP at the moment for the necessary feature discovery or enablement, but the underlying information is available, so wiring up some means of accessing it wouldn't be an enormous change. This might be through implementation of existing timestamping socket options, use of the existing XDP_OPTIONS socket option, or something else. Being able to determine whether the device supports HW timestamps also feels like a generally sane and useful thing. I don't have the context of any discussions that happened around adding HW timestamp support to AF_XDP, but perhaps you have a feel for whether there's reasons why this might or might not be feasible to add to the kernel AF_XDP support?

wdebruij:
API for feature discovery under review in upstream Linux:

https://lore.kernel.org/bpf/20230908225807.1780455-1-sdf@google.com/T/#m26ab89ea0b65b9ff92efbd95360703bfcfc20f91

This is like the XDP_OPTIONS suggestion. But because these are XDP features, not AF_XDP, and it is common to query using netlink, that is the approach.

This draft Onload AF_XDP change can easily be revised to use that query interface.

wdebruij:
The upstream XDP feature discovery series has been accepted, in time for Linux v6.7.

I will revise this series to use that. (will take a few weeks, probably)

wdebruij:
Apologies for the delay.

With the new device interface to query xdp features, the entire xdp option table can be replaced with just a few lines in __af_xdp_nic_init_hardware:

```
@@ -1085,6 +1085,10 @@ has_map_and_bound_prog:

        af_xdp_nic_tweak_hardware(nic);

+       if (nic->net_dev->xdp_metadata_ops &&
+           nic->net_dev->xdp_metadata_ops->xmo_rx_timestamp)
+               nic->flags |= NIC_FLAG_HW_RX_TIMESTAMPING;
+
```

I'll do that and update this pull request.

wdebruij:
> I'll do that and update this pull request.

Done. The code has gotten simpler without the option table.

Tested on Google cloud with a Debian image with Linux 6.3, with a Google gve.ko driver with mock support for receive timestamps in xdp (patch below).

This can probably also be tested on any 6.3+ kernel with veth, which supports this API.

```
@@ -2075,6 +2074,19 @@ static void gve_write_version(u8 __iomem *driver_version_register)

+static int gve_xdp_rx_timestamp(const struct xdp_md *ctx, u64 *timestamp)
+{
+       const u64 nsec_per_sec = 1000ULL * 1000 * 1000;
+       static u64 gve_tstamp = nsec_per_sec;
+
+       *timestamp = gve_tstamp++;
+       return 0;
+}
+
+static const struct xdp_metadata_ops gve_metadata_ops = {
+       .xmo_rx_timestamp = gve_xdp_rx_timestamp,
+};
+
 static int gve_probe(struct pci_dev *pdev, const struct pci_device_id *ent)

@@ -2137,6 +2149,9 @@ static int gve_probe(struct pci_dev *pdev, const struct pci_device_id *ent)
        dev->ethtool_ops = &gve_ethtool_ops;
        dev->netdev_ops = &gve_netdev_ops;
 
+       dev->xdp_metadata_ops = &gve_metadata_ops;
```

sianj-xilinx:
Thanks very much for the updates. I'm sorry, I'm a bit busy right now, so it'll be next week before I have a chance to look through it, but definitely looks nicer without the option table!

sianj-xilinx:
@wdebruij I'd like to apologise for still not having managed to review this. I haven't been working on AF_XDP related code recently, and every time this makes it close to the top of my list I find that I don't have time to context switch that state back in sufficiently to give this a proper review. I'm about to go on holiday now, so I'm afraid I'm going to continue to fail to review it for even longer, but hope to address it in the new year unless another member of the team gets there before me.

wdebruij:
Thanks for the heads-up, @sianj-xilinx. Good to know this is still on your radar.

Fully understand that it's not your priority. And it's a fairly significant change too.

wdebruij:
Friendly ping.

I reviewed the code myself to refresh my memory too.

Most lines of code are not Onload strictly, but helpers: patch 3 and 4 add an optional XDP loader and example XDP program.

Patch 1 allows replacing the default XDP program.
    This may be useful independent of the timestamp feature.

    For one, I've been looking at supporting XDP without having to disable RSS / insert NTuple rules.
    This recent commit allows receiving packets over multiple queues as long as they share the xdm umem:
    https://lore.kernel.org/all/65c0f032ac71a_7396029419@willemb.c.googlers.com.notmuch/T/

Patch 2 advertises NIC_FLAG_HW_RX_TIMESTAMPING iff the device supports XDP rx timestamping
Patch 5 reads the timestamp

    These two are hopefully small (22 LoC) and straightforward.

Let me know if there is anything I can do to help make it easier to review. Thanks!

ivatet-amd:
Hi @wdebruij, and sorry for the delay! I will take a look at this patch series early next week (Monday or Tuesday) unless anyone else does it sooner. 👍

wdebruij:
Awesome, thanks @ivatet-amd

wdebruij:
I just noticed that I need to explicitly include net/xdp.h in src/lib/efhw/af_xdp.c after upstream

"[PATCH bpf-next v2 3/3] net: invert the netdevice.h vs xdp.h dependency"
https://lore.kernel.org/all/20230803010230.1755386-4-kuba@kernel.org/

Which landed in v6.6.

Will update the pull request with that one line change, and rebase this series on top of this repo's master at the same time.

With this change it builds fine for me on Debian trixie with kernel 6.6.15-2

src/tools/bpf_link_helper/xdp_onload_prepare.c
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: GPL-2.0
+
+/* xdp_onload_prepare -- load an XDP ELF object with xsk map and program
+ *
+ * cd $SOMEDIR
+ * git clone --recurse-submodules https://github.com/libbpf/bpftool.git
+ * cd bpftool
+ * export BPFTOOL=$PWD
+ * make
```
ivatet-amd, line 9:
```suggestion
 * make -C libbpf/src
```
There's no Makefile in the top directory, hence "diving" into `libbpf/src`. I haven't yet verified, but `make -C src` could be a better one! The version I've checked out is https://github.com/libbpf/bpftool/commit/22621e39da611c292b7c13d89af7bdccc7b905fb.

src/tools/bpf_link_helper/xdp_tstamp.c
```diff
@@ -0,0 +1,89 @@
...
+ *
+ *
+ * # BUILD
+ *
+ * # install bpftool (if not installed yet)
+ * cd $SOMEDIR
+ * git clone --recurse-submodules https://github.com/libbpf/bpftool.git
+ * cd bpftool
+ * export BPFTOOL=$PWD
+ * make
```
ivatet-amd, line 15: Ditto.

src/tools/bpf_link_helper/xdp_tstamp.c
```diff
@@ -0,0 +1,89 @@
...
+ * cd $SOMEDIR
+ * git clone https://github.com/xdp-project/xdp-tools.git
+ * cd xdp-tools
+ * XDPTOOLS=$PWD
+ * ./configure
+ * make
+ *
+ * cd $ONLOAD
+ * cd src/tools/bpf_link_helper
+ * clang -g -target bpf -O2 -I $BPFTOOL/src/libbpf/include -I XDPTOOLS/headers -c xdp_tstamp.c -o ./xdp_tstamp.o
```
ivatet-amd, line 27:
```suggestion
 * clang -g -target bpf -O2 -I $BPFTOOL/libbpf/src -I $XDPTOOLS/headers -c xdp_tstamp.c -o ./xdp_tstamp.o
```

src/tools/bpf_link_helper/xdp_tstamp.c
```diff
@@ -0,0 +1,89 @@
...
+ *
+ *
+ * # RUN
+ *
+ * # insert the xdp prog *before* loading the onload kernel modules
+ * ./xdp_onload_prepare $DEV xdp_tstamp.o
+ */
+
+#include <linux/bpf.h>
+#include <bpf/bpf_helpers.h>
```
ivatet-amd, line 37:
```suggestion
#include <bpf_helpers.h>
```

ivatet-amd (commented):
@wdebruij I'm giving it a go, and I thought it would be a good idea to post some notes in the interim. If they look good to you, I'll apply them to the corresponding patches in the series.

src/tools/bpf_link_helper/xdp_onload_prepare.c
wdebruij: Thanks @ivatet-amd .

I must have messed up the comments. These changes look good to me.

Come to think of it, if the only thing needed from bpftool is libbpf, then it may be cleaner to just add that dependency (https://github.com/libbpf/libbpf).

+1 on the suggested changes. If we want to change more thoroughly, I can do that too.

ivatet-amd: Thanks @wdebruij. It seems the `libbpf` alone is sufficient. No worries, I'll apply it too. 👍

wdebruij:
Based on your latest update, I took another look at why I bothered with adding a new xdp_onload_prepare utility, which adds that libbpf dependency.

I do not immediately see a hard requirement for it for the rx timestamps. The way existing src/lib/efhw/af_xdp_bpf.c approach, to have a prepared BPF code program in xdp_prog_load, can perhaps be extended to include the new bpf_xdp_adjust_meta and bpf_xdp_metadata_rx_timestamp operations. That last operation is non-obvious in the kernel in that it is a driver-specific extension. But that complexity is dealt with by the BPF validator.

Aside: as user, being able to bypass the default program and insert my own XDP program is useful. Patch 1 optionally enables this. I did not mean to introduce hard external dependencies. Besides bpftool, quickly this also includes clang and as you pointed out a BTF capable kernel build (although I think that all distro kernels from ~2020 to have this, e.g., RHEL 8.2+)

ivatet-amd:
Yes, the user's ability to reload the XDP program is the perfect reason for a new utility. I also want to try https://github.com/xdp-project/xdp-tools/tree/master/xdp-loader when we get the infra issue sorted.

Yes, you're right about `CONFIG_DEBUG_INFO_BTF`. My latest attempt was in RHEL 9.3, and it was unfortunate that the config I used for v6.7 (with `xdp_metadata_ops`) had it commented out.

Please stand by. This is at the top of my task backlog.

ivatet-amd (approved):
(Limited) tested in Ubuntu 23.10 with Linux 6.5.

I want to simplify the instructions slightly and suggest installing the `libbpf-dev` and `libxdp-dev` packages, with the option to build them from sources for users of other OS vendors to avoid fiddling with the env variables and GCC and clang invocations.

I couldn't tame `xdp-loader` to replace the `xdp_onload_prepare` utility because it want to install the dispatcher app. Also, didn't have much success with `bpftool` and `ip` to do it either. Let's revisit it when we decide to put more effort into AF_XDP.

Scheduled it for the merge on Monday.

ivatet-amd:
@wdebruij It has been merged with minor modifications. Please let us know if something does not look right in the merged patch series. Thank you!

wdebruij:
Wonderful news! Thanks @ivatet-amd.

git diff compared to my previous version cherry-picked on top of 99a73d85 looks great. Improved integration and documentation.

PR #169, opened 2023-09-07
Fix build for Linux 6.5

ol-sergeyn:
linux-6.5: support updated pin_user_pages  
linux-6.5: compat for splice() and f_ops->sendpage

---
Build passes on Debian-testing (trixie) with 6.5 and 6.4 kernels.
Also checked with the related socket-tester tests:
```
--tester-run=sockapi-ts/ --ool=onload --tester-req=SENDFILE
--tester-run=sockapi-ts/usecases/splice:splice_before_data=FALSE,set_move=FALSE,set_nonblock=FALSE,extra_pipes=0,overfill_pipe=FALSE,acc_pipe=FALSE,di
ff_stacks=FALSE,check_after_splice=FALSE --ool=onload
```
No regressions.

### DIFF
--- a/src/lib/efthrm/tcp_helper_linux.c
+++ b/src/lib/efthrm/tcp_helper_linux.c
@@ -873,6 +873,230 @@ static unsigned linux_tcp_helper_fop_poll_alien(struct file* filp,
   return alien_file->f_op->poll(alien_file, wait);
 }
 
+#ifndef EFRM_HAVE_FOP_SENDPAGE
+/* Linux >= 6.5 */
+#include <linux/splice.h>
+static int
+tcp_helper_bvec_sendmsg(ci_netif* ni, ci_tcp_state* ts, struct msghdr *msg)
+{
+  int i, ret;
+  struct iovec *io = CI_ALLOC_ARRAY(typeof(*io), msg->msg_iter.nr_segs);
+
+  if( io == NULL )
+    return -ENOMEM;
+
+  for( i = 0; i < msg->msg_iter.nr_segs; ++i ) {
+    const struct bio_vec *bvec = &msg->msg_iter.bvec[i];
+
+    io[i].iov_base = (char*)kmap(bvec->bv_page) + bvec->bv_offset;
+    io[i].iov_len = bvec->bv_len;
+  }
+
+  ret = ci_tcp_sendmsg(ni, ts, io, msg->msg_iter.nr_segs,
+                        (ts->s.b.sb_aflags & CI_SB_AFLAG_O_NONBLOCK) | msg->msg_flags,
+                        CI_ADDR_SPC_KERNEL);
+  ci_free(io);
+  return ret;
+}
+
+/* X-SPDX-Source-URL: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git */
+/* X-SPDX-Source-Tag: v6.5 */
+/* X-SPDX-Source-File: fs/pipe.c */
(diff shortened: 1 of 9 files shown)


ol-alexandra (approved):

jbondpre-amd (approved):
relevant Linux commits:
https://github.com/torvalds/linux/commit/0b295316b3a9b7858eafbebdc31b4827a6edde03
https://github.com/torvalds/linux/commit/2dc334f1a63a8839b88483a3e73c0f27c9c1791c

src/lib/efthrm/tcp_helper_linux.c
```diff
@@ -873,6 +873,230 @@ static unsigned linux_tcp_helper_fop_poll_alien(struct file* filp,
...
+ *                 Space indented lines are Onload related code.
+ *                 'ci' prefix is added in order not to clash with Linux names.*/
+static void ci_wakeup_pipe_writers(struct pipe_inode_info *pipe)
+{
+	smp_mb();
+	if (waitqueue_active(&pipe->wr_wait))
+		wake_up_interruptible(&pipe->wr_wait);
+	kill_fasync(&pipe->fasync_writers, SIGIO, POLL_OUT);
+}
+static ssize_t ci_splice_to_socket(struct pipe_inode_info *pipe, struct file *out,
```
rhughes-xilinx, line 943: So, for my own thought process, here are the options:
1. This. Lots of kernel code copied, with the usual risks of divergence and compat problems. In particular the promise that this code will change again with folios
2. Use `__splice_from_pipe()`. The loop push-down promise will never be achieved with Onload, but we don't treat splice as a fast-path anyway
3. Extract `os_sock->f_ops->splice_write` from the backing socket, and call that (with appropriate munging of inputs). Goes through additional layers for Onload sockets, and potentially subject to compat problems if the kernel does something nuts. Also might not work at all

This is all fairly elementary code so I'm not opposed to option (1), but I'm also not convinced that one of the slower options wouldn't be preferable. In summary, I've argued myself in a circle again.

ol-sergeyn: Thank you for the review.
The option (2) was my initial approach, but then I decided it is better to follow new Linux-6.5 way and use `splice_to_socket`, keeping the old code (for linux-6.4 and older) untouched.

ol-sergeyn:
> relevant Linux commits: [torvalds/linux@0b29531](https://github.com/torvalds/linux/commit/0b295316b3a9b7858eafbebdc31b4827a6edde03) [torvalds/linux@2dc334f](https://github.com/torvalds/linux/commit/2dc334f1a63a8839b88483a3e73c0f27c9c1791c)

Thanks a lot for noting that!
I forgot to mention these commits in the PR description.

src/lib/efthrm/tcp_helper_linux.c
ivatet-amd: Shall we commit to this approach now and see if we can do any better when/if another kompat issue arises, e.g. to accommodate folios?

(2) seems to change the semantics a little, and (3) has more moving parts and therefore seems scary, but maybe next time, it will be more apparent if one of these approaches is better.

I suppose we could also try exporting `splice_to_socket()` and upstreaming it, but Linux 6.5 and 6.6 are already here, and I can't think of an appealing justification for making it an exported symbol.

ivatet-amd (approved):

rhughes-xilinx (approved):

PR #171, opened 2023-09-18
transport/ip: re-check TCP state when spinning

ol-alexandra:
When spinning in sendmsg(), Onload used to re-check TCP state for changes after poll only.  It is wrong, because poll may happen in another context.  Particularly, for TCP loopback connection we always have another thread/process which is changing thi TCP state.

----
In some cases an Onload stack may be polled by other process (or by kernel).  It is important that a process spinning in TCP send() accepts such changes in time.

No regression in Socket Tester sanity check:
`./run.sh -q --tester-run=sockapi-ts/ --ool=onload --tester-req=V5_SANITY --ool=spin --ool=loop4`

### DIFF
--- a/src/lib/transport/ip/tcp_send.c
+++ b/src/lib/transport/ip/tcp_send.c
@@ -1617,21 +1617,20 @@ ci_inline int ci_tcp_sendmsg_spin(ci_netif* ni, ci_tcp_state* ts,
 
   do {
     if( ci_netif_may_poll(ni) ) {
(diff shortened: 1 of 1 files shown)


ivatet-amd (approved):

ivatet-amd:
Thank you, Alexandra. 👍

Smoke-tested on dev DUTs with `4.18.0-348.el8.x86_64`, merged to `master`, cherry-picked to `onload-8.1`.

PR #175, opened 2023-09-26
Lg/on 15190

ligallag-amd:
This PR addresses the ON-15190 ticket. 
## Problem to be fixed
Running `EF_CTPIO=0 onload interactive -c "socket udp; bind $MCAST 5555 & "` multiple times will cause too many stacks to be created and onload will run out. This is recreated with:
``` shell
for i in  {5555..5575} ; do EF_CTPIO=0 onload interactive -c "socket udp; bind $MCAST $i" & done
```
Then when running with `EF_CTPIO=1`  a stack will not be able to be created and the resulting error is given:
```
netif_tcp_helper_alloc_u: ERROR: Failed to allocate stack (rc=-16)
See kernel messages in dmesg or /var/log/syslog for more details of this failure
```
## Outcome
In summary, when running onload with environment variable: `EF_CTPIO=0` on an X3 NIC it will now fail and will no longer create any resources. when running:
``` shell
for i in  {5555..5575} ; do EF_CTPIO=0 onload interactive -c "socket udp; bind $MCAST $i" & done
```
and outputs a message which can be viewed in dmesg. To confirm resources aren't taken, running with `EF_CTPIO=1` now succeeds while the `interactive` processes are are


src/lib/efthrm/tcp_helper_resource.c
```diff
@@ -1340,7 +1340,13 @@ get_vi_settings(ci_netif* ni, struct efhw_nic* nic,
   }
 
 #if CI_CFG_CTPIO
+  get_if_name(ni, info->intf_i, if_name);
   if( should_try_ctpio(ni, nic, info) ) {
+    if( NI_OPTS(ni).ctpio == 0 && nic->flags & NIC_FLAG_CTPIO_ONLY) {
```
ivatet-amd, line 1345:
```suggestion
    if( NI_OPTS(ni).ctpio == 0 && nic->flags & NIC_FLAG_CTPIO_ONLY ) {
```

ivatet-amd (approved):

PR #176, opened 2023-09-26
driver/linux_onload: remove wrong sendpage assertion

### LINKED ISSUE #173: assertion in splice()
I'm using `splice()` in following way: OS socket -> OS pipe -> Onload socket.  Sometimes I see following assertion:
```
[11220.108079] [onload] ci_assert((offset + size)<=(4096))
[11220.114533] [onload] where [offset + size=100c]
[11220.120205] [onload] and [4096=1000] 
[11220.124798] [onload] at /var/tmp/session_2022.05.07/onload-6840772e686a4c345e662596ff14e43d98cdb759-5.14.0-70.el9.x86_64-debug/build/x86_64_linux-5.14.0-70.el9.x86_64/driver/linux_onload/tcp_sendpage.c:57 
[11220.146305] [onload] FAIL at /var/tmp/session_2022.05.07/onload-6840772e686a4c345e662596ff14e43d98cdb759-5.14.0-70.el9.x86_64-debug/build/x86_64_linux-5.14.0-70.el9.x86_64/driver/linux_onload/tcp_sendpage.c:57 
[11220.168316] [onload] from /var/tmp/session_2022.05.07/onload-6840772e686a4c345e662596ff14e43d98cdb759-5.14.0-70.el9.x86_64-debug/build/x86_64_linux-5.14.0-70.el9.x86_64/driver/linux_onload/tcp_sendpage.c:57
```
I see this issue with linux>=6.1 and with recent CentOS kernels.
Everything works with NDEBUG build, or with DEBUG build when I remove this assertion.

I have an explanation.  Modern network drivers (including SFC) use compound pages as receive buffers; these big pages are split into smaller buffers of MTU size (mtu + some headers, etc ).  As a result `skb->data` of a received packet is continuous, but it may cross the bounds or an ordinary page.  This chunk of data is passed to OS pipe, and then to `linux_tcp_helper_fop_sendpage()`, where the assertion is fired.

I'll propose a patch removing the guilty assertion.

Please feel free to comment my considerations above.

ol-alexandra:
Onload used to assert that a memory chunk passed to sendpage handler is always within one page.  It was true at the time when the assertion was added.  It is not true any more.

If a user runs two splice() calls: OS socket -> OS pipe -> Onload socket, then Onload's sendpage handler gets a whole memory-continuous packet as it was received by the NIC.  Such a packet tend to be a part of a compound page, and it may cross the bounds of an ordinary page.

The real sense of this assertion was "ensure that we are handling a continuous chunk".  This is still true.  But the assertion is bad for linux>=6.1 and for some updates of older kernels.

Fixes #173

### DIFF
--- a/src/driver/linux_onload/tcp_sendpage.c
+++ b/src/driver/linux_onload/tcp_sendpage.c
@@ -54,7 +54,6 @@ ssize_t linux_tcp_helper_fop_sendpage(struct file* filp, struct page* page,
   ci_assert(page);
   ci_assert_ge(offset, 0);
   ci_assert_gt(size, 0);
-  ci_assert_le(offset + size, CI_PAGE_SIZE);
 
 #ifndef MSG_SENDPAGE_NOTLAST
   /* "flags" is really "more".  Convert it. */


src/driver/linux_onload/tcp_sendpage.c
```diff
@@ -54,7 +54,6 @@ ssize_t linux_tcp_helper_fop_sendpage(struct file* filp, struct page* page,
   ci_assert(page);
   ci_assert_ge(offset, 0);
   ci_assert_gt(size, 0);
-  ci_assert_le(offset + size, CI_PAGE_SIZE);
```
ivatet-amd, line 57: Would it be helpful to keep the assert if the page is huge/THP? e.g.
```
ci_assert((offset + size <= CI_PAGE_SIZE) || PageCompound(page) || PageHuge(page));
```

ivatet-amd (commented):
Sorry for the delay in reviews. I'm planning to take a look and merge https://github.com/Xilinx-CNS/onload/pull/178 and https://github.com/Xilinx-CNS/onload/pull/182 next week.

src/driver/linux_onload/tcp_sendpage.c
ol-alexandra: No.  `PageCompound(page)` is true for the first "small" page of the compound; it is false for all the following pages.

ivatet-amd: > `PageCompound(page)` is true for the first "small" page of the compound; it is false for all the following pages.

Huh, that's interesting! I guess it's a newer kernel feature.

Could you try running https://gist.github.com/ivatet-amd/68a808e2e571871f211c145157fc8825? In my older kernels, it shows the THP and HUGE bits set across the entire huge or TH page:
```
$ echo 10 | sudo tee /proc/sys/vm/nr_hugepages
$ sudo ./test.elf
THP
pagemap: 0x81800000000c6400 kpageflags: 0x40d868 HEAD THP
pagemap: 0x81800000000c6401 kpageflags: 0x411800 TAIL THP
pagemap: 0x81800000000c6402 kpageflags: 0x411800 TAIL THP
pagemap: 0x81800000000c6403 kpageflags: 0x411800 TAIL THP
Hugepage
pagemap: 0x818000000009d000 kpageflags: 0x29818 HEAD HUGE
pagemap: 0x818000000009d001 kpageflags: 0x800031800 TAIL HUGE
pagemap: 0x818000000009d002 kpageflags: 0x31800 TAIL HUGE
pagemap: 0x818000000009d003 kpageflags: 0x31800 TAIL HUGE
```

I believe `PageCompound()` would have returned "true" in the kernel space. Does it return "false" and break down the assertion?

ivatet-amd: >  I'm unable to reproduce this issue now.

That's fine. My point was about relaxing the assertion rather than dropping it entirely.

Would it work for you if (1) we merge the suggested patch, and (2) if the repro happens again with the updated assertion, we drop it?

ol-alexandra: I believe that you run Socket Tester, and I guess it happens to show the issue from time to time.  So it seems reasonable to apply a modified patch and look at the test results.

ivatet-amd: I've checked with @rhughes-xilinx, and we agreed to drop the assertion, considering `sendpage` is dropped from FOPs in the latest kernel/has a limited lifespan.

ivatet-amd (approved):

ivatet-amd:
Merged at https://github.com/Xilinx-CNS/onload/commit/9ef9f19916a6bd6182db84138eecd063b17e3f58.

PR #178, opened 2023-10-10
lib/efthrm: fix packet injection to kernel

ol-alexandra:
Ensure that a list of packets to be injected to kernel is properly detached from the stack structures, so that it is not updated under the feet of the deferred work.

Also add a guard to avoid dead loop in kernel when shared state is corrupted.

--------

Socket Tester with `V5_SANITY` is still happy.

Unfortunately I do not have a reproducer for the potential issue I'm fixing here.  To reproduce the issue, we need a stream of incoming packets which we'd like to inject to kernel.  If it is a TCP stream, then TCP protocol would repair any drops.  If it is a UDP stream, then Linux is allowed to drop the packets if it can't handle them all (ST really sees this).  There should be some stream rate which we are able to handle without loss, but it is not an easy thing to test.

May be the boss of Socket Tester @okt-yurijp has any ideas.

### DIFF
--- a/src/lib/efthrm/tcp_helper_resource.c
+++ b/src/lib/efthrm/tcp_helper_resource.c
@@ -8401,7 +8401,11 @@ corrupted:
 struct oo_inject_packets_work_data {
   struct work_struct work;
   tcp_helper_resource_t* trs;
+
+  /* This is kernel_packets_* field from the stack state,
+   * detached from the stack itself under the stack lock. */
   oo_pkt_p pkt_head;
+  unsigned n_pkts;
 };
 
 static void oo_inject_packets_work(struct work_struct* work)
@@ -8411,9 +8415,12 @@ static void oo_inject_packets_work(struct work_struct* work)
   ci_netif* ni = &data->trs->netif;
   ci_ip_pkt_fmt* pkt;
   int netif_is_locked;
+  unsigned n;
 
-  /* Part one: inject all packets to the kernel */
-  for( pkt = PKT_CHK(ni, data->pkt_head); ; pkt = PKT_CHK(ni, pkt->next) ) {
+  /* Part one: inject all packets to the kernel without stack lock. */
+  for( pkt = PKT_CHK(ni, data->pkt_head), n = 0;
+       n < data->n_pkts;
+       pkt = PKT_CHK(ni, pkt->next), n++ ) {
     /* No need to check the return value here.  If the function fails, the
      * packet is dropped, and a counter is incremented. */
     oo_inject_packet_kernel(ni, pkt);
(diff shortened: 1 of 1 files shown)


src/lib/efthrm/tcp_helper_resource.c
```diff
@@ -8413,24 +8417,36 @@ static void oo_inject_packets_work(struct work_struct* work)
...
+       n < data->n_pks;
+       pkt = PKT_CHK(ni, pkt->next), n++ ) {
     /* No need to check the return value here.  If the function fails, the
      * packet is dropped, and a counter is incremented. */
     oo_inject_packet_kernel(ni, pkt);
 
     if( OO_PP_IS_NULL(pkt->next) )
       break;
+    ci_assert_lt(n, data->n_pks);
+    if( n == data->n_pks )
```
ivatet-amd, line 8433: I guess this would be a tighter check:
```suggestion
    if( n >= data->n_pks )
```
Could you clarify: when does it occur? `n` and `n_pks` both seem kernelspace-only variables and the for-loop statement prevents us from getting here.

Presumably, a list of `pkt` can be corrupted, but the `n < data->n_pks;` silently saves us?

ivatet-amd (commented):
Thank you. I'm genuinely trying to comprehend the patch and would appreciate it if you could review my questions/suggestions.

src/lib/efthrm/tcp_helper_resource.c
ol-alexandra: Yes, the list of packets can be corrupted (because the packets are UL-mapped), and we should limit the loop with something non-shared-with-UL.  That's one of the ideas of this patch.  Such an idea already happens in some places in Onload.

Relevant part of the commit log:
```
Also add a guard to avoid dead loop in kernel when shared state
is corrupted.
```

ivatet-amd: Thanks @ol-alexandra. Does the assertion and `if` ever occur though? Both `n` and `data->n_pks` are non-shared.

ol-alexandra: In theory: the last packet is corrupted , and `pkt->next` is non-null.  So we don't break at
```
   if( OO_PP_IS_NULL(pkt->next) )
      break;
```
because of corruption.  Then we hit assertion (assertions traditionally assume well-behaving UL) and the condition.  The condition definitely needs another break; I'll fix.

src/lib/efthrm/tcp_helper_resource.c
```diff
@@ -8411,24 +8415,40 @@ static void oo_inject_packets_work(struct work_struct* work)
...
   int netif_is_locked;
+  unsigned n;
 
   /* Part one: inject all packets to the kernel */
-  for( pkt = PKT_CHK(ni, data->pkt_head); ; pkt = PKT_CHK(ni, pkt->next) ) {
+  for( pkt = PKT_CHK(ni, data->pkt_head), n = 0;
+       n < data->n_pkts;
+       pkt = PKT_CHK(ni, pkt->next), n++ ) {
     /* No need to check the return value here.  If the function fails, the
      * packet is dropped, and a counter is incremented. */
```
rhughes-xilinx, line 8425: Can we fix this comment while we're here, please?

src/lib/efthrm/tcp_helper_resource.c
ivatet-amd: Thanks, that's how I understand it too, but what puzzles me is:
```c
for( ... , n = 0;
     n < data->n_pkts; /* We shouldn't enter the loop if n >= data->n_pkts. */
     ... , n++ ) {

  /* Nothing changes @n and @data (including @data->n_pkts). */

  /* I can't understand why this can occur even if UL fiddles with shared memory. */
  ci_assert_lt(n, data->n_pkts);
  if( n >= data->n_pkts ) {
    ...
    break;
  }
}
```

src/lib/efthrm/tcp_helper_resource.c
```diff
@@ -8422,17 +8429,31 @@ static void oo_inject_packets_work(struct work_struct* work)
       break;
   }
 
-  /* Part two: free Onload packets */
+  /* If the packet list was not corrupted, then we break from the above
+   * loop just before n becomes equal to data->n_pkts. */
+  ci_assert_equal(n + 1, data->n_pkts);
```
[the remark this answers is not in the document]
ivatet-amd: Now I understand, thank you. 👍

PR #183, opened 2023-10-30
tcp_helper_resource: configurable XDP headroom via module param

zhuyifei1999:
At Google we have a driver where it expects the posted buffers to be 128-byte aligned. However EF_VI_DMA_ALIGN only enforces 64-byte alignment and in testing we found that `offsetof(ci_ip_pkt_fmt, dma_start) == 192`.

This adds a module param such that the headroom is configurable. The minimum remains `offsetof(ci_ip_pkt_fmt, dma_start)`, and the maximum is having bytes remaining after subtracting the configured headroom plus XDP_PACKET_HEADROOM from CI_CFG_PKT_BUF_SIZE, and the configured headroom remain aligned by EF_VI_DMA_ALIGN to preserve any assumptions onload makes.

This way, we can perform a:
```
$ echo 256 > /sys/module/onload/parameters/xdp_headroom
```
in order to get an 128-byte alignment that our driver expects. Existing users of onload are not impacted because the default is the same as the original value before this change.

### DIFF
--- a/src/lib/efthrm/tcp_helper_resource.c
+++ b/src/lib/efthrm/tcp_helper_resource.c
@@ -129,6 +129,15 @@ MODULE_PARM_DESC(periodic_poll_skew,
                  "Allowed time skew for periodic polls.  "
                  "Defaults to 10ms.");
 
+unsigned int xdp_headroom = offsetof(ci_ip_pkt_fmt, dma_start);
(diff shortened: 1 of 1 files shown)


ivatet-amd (approved):
Sorry for the delay in the review!

PR #186, opened 2023-11-10
lib/transport: fix potential segfault in ci_rx_pkt_ifindex()

okt-sergeyn:
In ci_rx_pkt_ifindex() if pkt is multicast, l.mac is set to NULL and then passed to ci_log() leading to segfault.
The patch fixes it by passing non-NULL pointer to ci_log.


ol-alexandra (approved):

jfeather-amd (approved):
Looks like a reasonable change to me!

jfeather-amd:
Merging this change now, thank you!

PR #187, opened 2023-11-15
driver: fix out-of-bounds UBSAN warning

okt-sergeyn:
Since commit 2d47c6956ab3 ("ubsan: Tighten UBSAN_BOUNDS on GCC") in Linux-6.5 we can't use arrays of size [0] or [1] because it triggers UBSAN false-positive warnings "array-index-out-of-bounds".

Use __DECLARE_FLEX_ARRAY() macro to declare flexible arrays in those cases. For arrays with size [1] use a union to keep the size of a struct containing the flexible array unchanged.

---

Depends on #169
Checked on Ubuntu 23.10 `6.5.0-9-generic`

### DIFF
--- a/src/include/ci/tools/sysdep.h
+++ b/src/include/ci/tools/sysdep.h
@@ -62,6 +62,24 @@
 typedef ci_int32 ci_uerr_t; /* range of OS user-mode return codes */
 typedef ci_int32 ci_kerr_t; /* range of OS kernel-mode return codes */
 
+#ifndef __DECLARE_FLEX_ARRAY
+#ifndef __cplusplus
+/* linux < 5.16. */
+/* X-SPDX-Source-URL: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git */
+/* X-SPDX-Source-Tag: v5.16 */
+/* X-SPDX-Source-File: include/uapi/linux/stddef.h */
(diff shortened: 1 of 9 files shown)


src/include/ci/tools/sysdep.h
```diff
@@ -62,6 +62,22 @@
...
+/* X-SPDX-Source-URL: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git */
+/* X-SPDX-Source-Tag: v5.16 */
+/* X-SPDX-Source-File: include/uapi/linux/stddef.h */
+/* X-SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
+/* X-SPDX-Comment: Linux 5.16 got this macro to declare flexible arrays.
+ *                 Onload uses it. In order to keep compatibility with
+ *                 older kernels, place the macro here. */
+#define __DECLARE_FLEX_ARRAY(TYPE, NAME)	\
+	struct { \
+		struct { } __empty_ ## NAME; \
```
rhughes-xilinx, line 76: This header file ends up included in places like citools, so I'd feel slightly safer if this macro were wrapped in `#ifndef __cplusplus` just to ensure that we'd definitely spot anybody who might be about to shoot themselves in the foot.

ol-alexandra (approved):

ivatet-amd:
Internal note: I plan to cherry-pick the patch to Onload 8.1 after release.

PR #195, opened 2023-12-13
X3: Catchup applications on missed superbufs upon wakeup

ligallag-amd:
# The problem
There is a ringbuffer shared between kernel and userspace which is used to send superbufs (sbufs) from the kernel.  This ringbuf is of size 16 and so if an app is paused it cannot take any sbufs from it and no more can be posted -  this causes drops. Unfortunately, this ringbuf's size cannot be changed otherwise the ABI will be broken.

Similarly there is a free queue ringbuffer of the same size used to post buffers from the user side to the kernel.  The difference being that if this one is filled completely, the application will abort.

Because of this, an application cannot fall asleep for long without either aborting or dropping a large number of packets.

---------------------
# Proposed solution
In general, the way around this is to hold on to the buffers until the shared memory queues are free to be pushed to.
## Posting from userspace to kernelspace
The efct_rx_descriptor buffer contains an entry for each sbuf in an rxq and so a linked list of sbufs which are yet  to be posted to the kernel can be maintained. The next time a superbuf is freed, it will try to post sbufs in this list also.
## Posting from kernelspace to userspace
Superbufs are now kept around longer and only removed once each app has received it or the nic is about to run out of  donated superbufs. In the latter case, the sequence number of the apps will be advanced.

When an application is destroyed, the superbufs it donated will be taken away and so another sleeping application  will need to be skipped ahead.

In the case where no buffers can be freed (all received buffers have been posted to the shared memory - particularly true for  small rxqs) and a new one arrives through efct_buffer_start, ensure that on efct_buffer_end we free that same donated buffer so the nic doesn't starve.

--------------------------------------------
# Testing
All tests verify that at their end, all sbufs have been freed with: `cat /proc/meminfo | grep HugePages_`.  Tests with high throughput and multiple apps need to be run in release mode meaning onload is compiled with `NDEBUG=1` otherwise  there will be "no descriptor" drops on the NIC. 

It should be noted that applications can only fall back by at most the number of superbufs that have been donated  to x3-net. The number of superbufs donated by each application is given by its rxq size. The number of superbufs it  donates is given by `ceil((rxq_size * bytes_per_packet) /  (bytes_per_superbuf))` where `bytes_per_packet=2KiB` and `bytes_per_superbuf=2MiB`

Note that an app is not able to buffer completely up to it's rxq size some of these same buffers need to be  returned so the nic is not starved. 

### Well behaved apps
| Order | DUT1 | DUT2|
|------- |--------|------|
| 1 | `efsink enp1s0f1np3 udp:239.1.2.3:22000 &` | |
| 2 | `efsink enp1s0f1np3 udp:239.1.2.3:22000` | |
| 3 | | `efsend enp1s0f1np3 239.1.2.3 22000 -n 200000 -m 1024` |

All packets received for both efsinks

### Sleepy app
| Order | DUT1 | DUT2|
|------- |--------|------|
| 1 | `EF_VI_RXQ_SIZE=65536 efsink enp1s0f1np3 udp:239.1.2.3:22000` -> `Ctrl z` | |
| 2 | | `efsend enp1s0f1np3 239.1.2.3 22000 -n 200000 -m 1024` |
| 3 | `fg` | |

Dut1 output:
```
$ fg
EF_VI_RXQ_SIZE=65536 efsink enp1s0f1np3 udp:239.1.2.3:22000
         0                0                0
     41470              353            41470
     22337              190            63807
```

### Sleepy app - small rxq
| Order | DUT1 | DUT2|
|------- |--------|------|
| 1 | `EF_VI_RXQ_SIZE=512 efsink enp1s0f1np3 udp:239.1.2.3:22000` -> `Ctrl z` | |
| 2 | | `efsend enp1s0f1np3 239.1.2.3 22000 -n 200000 -m 1024` |
| 3 | `fg` | |

Dut1 output:
```
$ fg
EF_VI_RXQ_SIZE=512 efsink enp1s0f1np3 udp:239.1.2.3:22000
         0                0               16
      1017                8             1033
```

### Consume, fall behind and awake
| Order | DUT1 | DUT2|
|------- |--------|------|
| 1 |  `EF_VI_RXQ_SIZE=65536 efsink enp1s0f1np3 udp:239.1.2.3:22000` |  |
| 2 |  | `efsend enp1s0f1np3 239.1.2.3 22000 -n 1000000 -m 1024` |
| 3 | `Ctrl-z` | |
| 4 |  `EF_VI_RXQ_SIZE=65536 efsink enp1s0f1np3 udp:239.1.2.3:22000` | |
| 5 | | `efsend enp1s0f1np3 239.1.2.3 22000 -n 1000000 -m 1024` |
| 6 | `Ctrl-c` , `fg` | |

The efsink that was put to sleep received `1,000,000` packets before it was put to sleep.  When it was awoken at time `6` It caught up on some of the buffers it missed leaving it with a total of 1061503 packets.

### Large catchup
| Order | DUT1 | DUT2|
|------- |--------|------|
| 1 | Create 15 efsinks with: `EF_VI_RXQ_SIZE=65536 efsink enp1s0f1np3 udp:239.1.2.3:22000`  and put them in the background with `Ctrl-Z` | |
| 2 | | `efsend enp1s0f1np3 239.1.2.3 22000 -n 1000000 -m 1024` |
| 3 | `for i in {1..15}; do bg; done`| |

Each application catches up by `943,680` packets because a large number of superbufs were donated to the nic. The rate of catchup is slow at: `279Mb/s` due to `napi_poll` not being triggered frequently enough as there are no incoming packets on the nic. If there are incoming packets during catchup, the buffered superbufs will be posted much faster due to a more frequent `napi_poll`. If the same as above is done but packets are continuously sent when the applications wake up, the catchup is much faster and a bandwidth of `~17.5Gb/s` is observed. This is on a 10 Gigabit nic so the  `~7.5Gb/s` bandwidth increase is due to the buffered superbufs.


ligallag-amd:
I have pushed a new commit to this branch which should allow users to reload the onload kernel module without having to recompile their user space program. As such, there will be differing views of `CI_EFCT_MAX_SUPERBUFS` from user and kernel space due to this patch. This case is now handled in the kernel module without returning an error.

An alternative fix to this patch is to return `CI_EFCT_MAX_SUPERBUFS` to 512 and deal with the consequences of that. This is done [on this branch](https://github.com/ligallag-amd/onload/tree/efct_superbufs_simple_patch) if this is preferable for whatever reason. The obvious problem with the fix on this branch is that the user won't be able to use more than 512 sbufs.

src/lib/efrm/efrm_efct_rxq.c
```diff
@@ -206,7 +210,7 @@ int efrm_rxq_refresh(struct efrm_efct_rxq *rxq, unsigned long superbufs,
 		kfree(pages);
 		return rc;
 	}
-	for (i = 0; i < CI_EFCT_MAX_HUGEPAGES; ++i) {
+	for (i = 0; i < n_hugepages; ++i) {
```
rhughes-xilinx, line 213: This one can remain at `CI_EFCT_MAX_HUGEPAGES`

src/lib/efrm/efrm_efct_rxq.c
```diff
@@ -189,10 +189,14 @@ int efrm_rxq_refresh(struct efrm_efct_rxq *rxq, unsigned long superbufs,
 	struct xlnx_efct_hugepage spare_page = {};
 	size_t i;
 	int rc = 0;
+	unsigned n_hugepages = CI_MIN(max_superbufs, (unsigned)CI_EFCT_MAX_SUPERBUFS) /
+                         CI_EFCT_SUPERBUFS_PER_PAGE;
 
 	if (max_superbufs != CI_EFCT_MAX_SUPERBUFS) {
-		/* Could be supported without much difficulty, but no need for now */
-		return -EINVAL;
+		EFRM_TRACE("max_superbufs: %u passed in by user not equal to kernel's: %u. "
```
rhughes-xilinx, line 196: I think I'd like to be noisier in the event that it's definitely going to go wrong, i.e. when `pages` has a non-NULL entry beyond max_superbufs then log an error. The goal is to make such failures in the field more debuggable.

src/driver/linux_resource/efct_superbuf.c
```diff
@@ -77,17 +73,89 @@ static bool post_superbuf_to_app(struct efhw_nic_efct_rxq* q, struct efhw_efct_r
...
+static void update_oldest_app_seq(struct xlnx_efct_device *edev,
+                              struct xlnx_efct_client *client, int qid,
+                              struct efhw_nic_efct_rxq* q)
 {
+  /* Update the oldest_app_seq pointer. When it advances, drop references
+   * to the sbufs it went past. */
+  uint32_t min = q->sbufs.added;
   struct efhw_efct_rxq *app;
+  for( app = q->live_apps; app; app = app->next ) {
+    min = next_app_seq_min(app, q, min);
```
rhughes-xilinx, line 119: I'm concerned about drift in this loop, i.e. (truncating to a 16-bit world to make things easier to read) added=0x9000, app1=0x2000, app2=0xb000, app3=0x7000. That'd make min=0x7000 when going through this loop, which is clearly wrong. Need to think more about how that case is supposed to work and, more generally, what happens if an app is stopped for a very long time; it might be necessary to mark it as 'dead' some other way if it drops more than 2^31 behind. Given how long 2^31 is, I wouldn't be opposed to breaking such an app entirely and irrevocable, if that turns out to be easier to implement.

src/driver/linux_resource/efct_superbuf.c
```diff
@@ -77,17 +73,89 @@ static bool post_superbuf_to_app(struct efhw_nic_efct_rxq* q, struct efhw_efct_r
...
+{
+  /* Starting from removed, find and drop superbufs until one is freed
+   * or `q->sbufs.oldest_app_seq >= q->sbufs.removed`. There is no point
+   * in skipping past bufs the driver hasn't yet removed. */
+  uint16_t sbid_to_free;
+  do {
+    if( SEQ_GE(q->sbufs.oldest_app_seq, q->sbufs.removed) )
+      return;
+    sbid_to_free = q->sbufs.q[q->sbufs.oldest_app_seq++ % CI_ARRAY_SIZE(q->sbufs.q)].sbid;
+  } while(!drop_sbuf_ref(edev, client, qid, q, sbid_to_free));
```
rhughes-xilinx, line 157: whitespace

src/driver/linux_resource/efct_superbuf.c
```diff
@@ -121,10 +187,18 @@ static void reap_superbufs_from_apps(struct xlnx_efct_device *edev,
     struct efhw_efct_rxq *app = *pprev;
     if( app->destroy ) {
       int sbid;
+
       ci_bit_for_each_set(sbid, app->owns_superbuf, CI_EFCT_MAX_SUPERBUFS)
         finished_with_superbuf(edev, client, qid, q, app, sbid);
       EFHW_ASSERT(app->current_owned_superbufs == 0);
+
+      /* Now this app is destroyed, its donated sbufs are gone and so
+       * we must free sbufs until we have an apropriate amount. */
```
rhughes-xilinx, line 196: appropriate

src/lib/efhw/efct.c
```diff
@@ -61,17 +64,29 @@ efct_nic_rxq_bind(struct efhw_nic *nic, int qid, bool timestamp_req,
...
+    efct_unprovide_hugetlb_alloc();
+    return used_hugepages;
+  }
+
+  EFHW_ASSERT(used_hugepages <= CI_EFCT_MAX_HUGEPAGES);
+
+  if( n_hugepages + used_hugepages > CI_EFCT_MAX_HUGEPAGES ) {
     /* We'd fail later on in bind_rxq(), but only after we'd done a load of
      * shuffling-about and memory allocation. So let's have this early sanity
      * check as well */
```
rhughes-xilinx, line 81: Didn't we add this code because this comment is no longer completely true?

src/lib/efhw/efct.c
```diff
@@ -107,6 +122,39 @@ efct_get_hugepages(struct efhw_nic *nic, int hwqid,
...
+  ssize_t used;
+  int i, rc;
+
+  pages = kvzalloc(sizeof(pages[0]) * CI_EFCT_MAX_HUGEPAGES, GFP_KERNEL);
+  if( ! pages )
+    return -ENOMEM;
+
+  /* This call can return `EINVAL` for a few reasons, including the case where
+   * the provided `qid` is not bound to by `nic`. Instead of returning an error
+   * here, we should instead validly claim that no hugepages are being used. */
```
rhughes-xilinx, line 138: Did I complain about this before? Can you remember what I said?

src/include/ci/internal/seq.h
```diff
@@ -18,4 +18,4 @@
 #define SEQ_MIN(x, y)           (SEQ_LE(x, y) ? (x) : (y))
 #define SEQ_MAX(x, y)           (SEQ_LE(x, y) ? (y) : (x))
 
-#endif /* __CI_INTERNAL_SEQ_H__ */
\ No newline at end of file
+#endif /* __CI_INTERNAL_SEQ_H__ */
```
rhughes-xilinx, line 21: Put this in the earlier commit

src/lib/efrm/efrm_efct_rxq.c
```diff
@@ -325,10 +325,15 @@ int efrm_rxq_refresh(struct efrm_efct_rxq *rxq, unsigned long superbufs,
 	struct xlnx_efct_hugepage spare_page = {};
 	size_t i;
 	int rc = 0;
+	bool is_sbuf_err_logged = false;
+	unsigned n_hugepages = CI_MIN(max_superbufs, (unsigned)CI_EFCT_MAX_SUPERBUFS) /
+	                       CI_EFCT_SUPERBUFS_PER_PAGE;
 
 	if (max_superbufs != CI_EFCT_MAX_SUPERBUFS) {
```
rhughes-xilinx, line 332: Lets make this `max_superbufs < CI_EFCT_MAX_SUPERBUFS`

rhughes-xilinx (commented):
I nearly approved of this. You might get away with fix-and-ship

src/lib/efhw/efct.c
ligallag-amd: I think we added this code so we don't end up donating superbufs to the pool such that an sbid `>= CI_EFCT_MAX_SUPERBUFS` isn't returned. I'll update the comment.

src/lib/efhw/efct.c
jfeather-amd: I think you might have previously, but I believe I addressed the issues you raised. I have a feeling it was because I made certain assumptions about what the X3 driver does if `rc == -EINVAL` and just fell through to the for loop because the allocation would still be valid (and zeroed) resulting in `used = 0`. Instead, we just return 0 early instead.

src/driver/linux_resource/efct_superbuf.c
ligallag-amd: Just to put we discussed into writing:
`q->sbufs.added - q->sbufs.oldest_app_seq <= CI_EFCT_MAX_SUPERBUFS << 2^31`
`next_app_seq_min` is called for each application each time an sbuf is posted to an app or when an app has finished with a superbuf and we succeed in posting another.
This function updates `app->next_sbuf_seq` to `q->sbufs.oldest_app_seq` whenever it is less than the oldest seq. 
This means the app's sequence numbers will never fall behind far enough to cause a drift.

ligallag-amd:
Commits have been merged into master so closing this PR:
0c4ee8983eb3c51792574c8fc637bb282be59229
7aae294bc2034af0105bd310f3cd4b9197c97fdb
66cdbd6d055514875c6bba6a0e04c33d6676000f
38130d26274ec04020d2ab08586b3225da489199

PR #197, opened 2024-01-11
onload_profiles: fix nginx profile to parser for -s option

ol-adubovoj:


ol-alexandra (approved):

scripts/onload_profiles/wait-old-cluster-exit.opf-fragment
```diff
@@ -5,7 +5,7 @@
 # ceased to exist
 # If references to old onload stacks are still being released - wait.
 
-if echo "$@" | perl -n -e 'if(/\s-s/) {exit 1}'; then
+if echo "$@" | grep -q -e '[[:space:]]-s'; then
```
ivatet-amd, line 8: I'm not familiar with the Nginx profile, please help me to clarify:

As-is, if the user provided `-s signal` to `nginx`, we meant to invoke `onload_stackdump`, but we didn't and needed to invert the condition.

If that's the case, may I suggest expanding the commit message and then let's merge it?

ivatet-amd: Also, `grep` looks neater than `perl`!

ol-alexandra: > As-is, if the user provided `-s signal` to `nginx`, we meant to invoke `onload_stackdump`, but we didn't and needed to invert the condition.

exactly

ol-adubovoj: I just expanded the commit-message.
Please check this out.

ol-adubovoj: @ivatet-amd are you happy with my extended commit message or can you suggest something else?

ivatet-amd: Awesome, thank you. 👍

ivatet-amd (approved):

PR #198, opened 2024-01-23
Refactor tcp/udp checksum calculating

okt-sergeyn:
Refactor tcp/udp checksum calculating and follow some RFC recommendations.

The main purpose of the patch series is to make Onload checksum calculations similar to Linux for the case of 0xffff tcp checksum. Onload calculates it as 0.
`ci_analyse_pkt()` output for such packets is:
`[onload] TCP ***** bad checksum ffff (I get 0) ***** `

Refactor checksum API and unify the callers to call directly ef_* functions instead of ci_* wrappers.
This also brings in the `include/etherfabric` API change. As I understand it is a kind of public API, so I'd appreciate any comments about that.

Fix `ci_{tcp,udp}_csum_correct()` functions to handle ffff checksum.
Fix tcp tx checksum computing to replace 0 with ffff (follow udp recommendations).

Add a simple unit test for ef_tcp_checksum().


ol-alexandra (approved):

src/lib/ciul/checksum.c
```diff
@@ -295,6 +295,61 @@ uint32_t ef_tcp_checksum_ipx(int af, const void* ipx, const struct tcphdr* tcp,
     return ef_tcp_checksum((const struct iphdr*)ipx, tcp, iov, iovlen);
 }
 
+int ef_udp_checksum_is_correct(const struct iphdr* ip, const struct udphdr* udp,
+                               const struct iovec* iov, int iovlen)
+{
+  uint64_t csum64;
+  uint32_t csum;
+
+  csum64 = (uint64_t)ip->saddr + ip->daddr +    /* This is the UDP */
```
rhughes-xilinx, line 312: UDPv4 packets are allowed to omit their checksum. You either need to handle that case here, or call out in the documentation that the caller needs to handle it.

src/lib/ciul/checksum.c
```diff
@@ -295,6 +295,61 @@ uint32_t ef_tcp_checksum_ipx(int af, const void* ipx, const struct tcphdr* tcp,
...
+           htons(IPPROTO_UDP) + udp->len;       /* pseudo-header */
+  csum64 = ip_csum64_partial(csum64, udp, sizeof(*udp));
+  csum64 = ip_csum64_partialv(csum64, iov, iovlen);
+  csum = ip_proto_csum64_finish(csum64);
+
+  return csum == 0;
```
rhughes-xilinx, line 318: My checksum-fu isn't great, but I thought this was supposed to be `csum == 0 || csum == 0xffff`, i.e. ==+0 or ==-0. Can you test this code with a packet whose checksum is actually 0 (so the sender will have negated it)?

okt-sergeyn: Please check out the `src/tests/unit/lib/ciul/checksum.c`, which I've added.
The test checks `ef_tcp_checksum_is_correct()` (I think udp is the same) with a packet which tcp checksum is ffff, i.e, it is initially 0 but Linux inverts it (Onload would compute it as 0 for that packet).  The test shows that both 0 and ffff for that packet are treated as good checksum. So I guess that `csum == 0` is enough.

rhughes-xilinx: If I change
```
  char tcpdata[] = {
    0xa9, 0xf6, 0x52, 0x13,
```
to
```
  char tcpdata[] = {
    0xa9, 0xf6, 0x52, 0x14,
```
then the test fails

rhughes-xilinx: Right, yes, I think I've convinced myself that +0 (before negation) is unachievable - thanks for your patience. Ones' complement is mind-bending.

src/include/etherfabric/checksum.h
```diff
@@ -97,6 +97,32 @@ extern uint32_t
...
+** \param payload       UDP payload.
+** \param payload_len   Length of the payload.
+**
+** \return The checksum of the UDP packet.
+**
+** Calculate the checksum for a UDP packet.  The UDP header must be populated
+** (with the exception of the checksum field itself, which is ignored) before
+** calling this function.
+*/
+ef_vi_inline uint32_t
```
rhughes-xilinx, line 114: Bit weird that you're returning u32 rather than u16, but it's consistent with the existing stuff, so OK

src/include/etherfabric/checksum.h
```diff
@@ -97,6 +97,32 @@ extern uint32_t
...
 
+/*! \brief Calculate the checksum for a UDP packet
+**
+** \param af            The address family of the IP header for the packet.
+** \param ipx           The IP header for the packet (struct iphdr or ipv6hdr).
+** \param udp           The UDP header for the packet.
+** \param payload       UDP payload.
+** \param payload_len   Length of the payload.
+**
+** \return The checksum of the UDP packet.
```
rhughes-xilinx, line 108: Please call out that the return value is already big endian

rhughes-xilinx (approved):

ivatet-amd:
I want to study this PR and merge it tomorrow morning.

ivatet-amd (approved):
Compatibility-wise, `src/include/etherfabric/checksum.h` must be good because it doesn't change the existing API or affect the kernel/user interface. Also, thank you for adding the checksum unit test. 👍

src/include/ci/net/ipvx.h
```diff
@@ -324,6 +324,12 @@ typedef union ci_ipx_hdr {
 #endif
 } ci_ipx_hdr_t;
 
+#if CI_CFG_IPV6
+#define ipx_hdr_ptr(af, ipx) (af == AF_INET6 ? &ipx->ip6 : &ipx->ip4)
```
ivatet-amd, line 328: I ran a quick test and discovered a built-time issue:
```console
$ TRANSPORT_CONFIG_OPT_HDR=ci/internal/transport_config_opt_cloud.h make -j16
...
src/include/ci/net/ipvx.h:328:58: error: pointer type mismatch in conditional expression [-Werror]
 #define ipx_hdr_ptr(af, ipx) (af == AF_INET6 ? &ipx->ip6 : &ipx->ip4)
                                                          ^
build/x86_64_linux-4.18.0-372.9.1.el8.x86_64/lib/transport/ip/netif_event.c:101:19: note: in expansion of macro ‘ipx_hdr_ptr’
   void *ipx_hdr = ipx_hdr_ptr(af, ipx);
                   ^~~~~~~~~~~
```

Could you apply the type-casting and see if it works for you?
```suggestion
#define ipx_hdr_ptr(af, ipx) \
    (af == AF_INET6 ? (void*)&ipx->ip6 : (void*)&ipx->ip4)
```

okt-sergeyn: Thanks! Fixed.

PR #199, opened 2024-01-23
mmake: fix LINUX_VERSION_3 variable for shorter versions

okt-sergeyn:
Use more flexible regular expression to handle versions which
contain 2 numbers, for instance 6.6-amd64.

Checked with 6.6, 6.5.0-5 and 5.14.0-163

### DIFF
--- a/mk/platform/linux.mk
+++ b/mk/platform/linux.mk
@@ -5,9 +5,15 @@ $(shell echo >&2 "KPATH is not set.")
 $(error KPATH is not set.)
 endif
 
+# Extract the Linux kernel version from a line read from utsrelease.h.
+linux_version = $(shell printf '$(1)' | sed -r 's/^#define UTS_RELEASE "([0-9]+(\.[0-9]+)*).*/\1/; t; d')
+
 LINUX		   := 1
 
-LINUX_VERSION_3	   := $(shell cat $(KPATH)/include/generated/utsrelease.h 2>/dev/null | sed 's/^\#define UTS_RELEASE \"\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/; t; d')
+LINUX_VERSION_N := $(call linux_version,$(file < $(KPATH)/include/generated/utsrelease.h))
+ifeq ($(LINUX_VERSION_N),)
+$(error Failed to extract kernel version from utsrelease.h)
+endif
 
 DRIVER		   := 1
 MMAKE_USE_KBUILD   := 1
@@ -43,6 +49,6 @@ greater_version = $(shell printf "$(1)\n$(2)\n" | sort --version-sort | tail -1)
 # This approach does not work on Linux < 5.4, so in order not to break old
 # kernels, do it only with Linux >= 6.0.
 KBUILD_LIB_MAKE_TRG ?=
-ifeq ($(call greater_version,$(LINUX_VERSION_3),6.0), $(LINUX_VERSION_3))
+ifeq ($(call greater_version,$(LINUX_VERSION_N),6.0), $(LINUX_VERSION_N))
 KBUILD_LIB_MAKE_TRG += $(lib_obj_path)
 endif


ol-alexandra (approved):

mk/platform/linux.mk
```diff
@@ -7,7 +7,7 @@ endif
 
 LINUX		   := 1
 
-LINUX_VERSION_3	   := $(shell cat $(KPATH)/include/generated/utsrelease.h 2>/dev/null | sed 's/^\#define UTS_RELEASE \"\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/; t; d')
+LINUX_VERSION_3 := $(shell grep -Eo '([0-9]+\.){1,}[0-9]+' $(KPATH)/include/generated/utsrelease.h 2>/dev/null)
```
ivatet-amd, line 10: I'm slightly worried that `grep` echoes a couple of matches for me, and I couldn't find out how to constrain it because both groups are in the same line, even thought it might be presented to the `greater_version()` separated with whitespace/not a newline:
```console
$ cat /lib/modules/$(uname -r)/build/include/generated/utsrelease.h
#define UTS_RELEASE "5.14.0-162.6.1.el9_1.x86_64"

$ cat /lib/modules/$(uname -r)/build/include/generated/utsrelease.h | sed 's/^\#define UTS_RELEASE \"\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/; t; d'
5.14.0

$ grep -Eo '([0-9]+\.){1,}[0-9]+' /lib/modules/$(uname -r)/build/include/generated/utsrelease.h
5.14.0
162.6.1
```

What do you think about this alternative?
```diff
diff --git a/mk/platform/linux.mk b/mk/platform/linux.mk
index ea71e96cfb..dceddd841a 100644
--- a/mk/platform/linux.mk
+++ b/mk/platform/linux.mk
@@ -5,9 +5,12 @@ $(shell echo >&2 "KPATH is not set.")
 $(error KPATH is not set.)
 endif

+# Extract the Linux kernel version from a line read from utsrelease.h.
+linux_version = $(shell sed -r 's/^#define UTS_RELEASE "([0-9]+(\.[0-9]+)*).*/\1/; t; d' <<< '$(1)')
+
 LINUX             := 1

-LINUX_VERSION_3           := $(shell cat $(KPATH)/include/generated/utsrelease.h 2>/dev/null | sed 's/^\#define UTS_RELEASE \"\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/; t; d')
+LINUX_VERSION_N           := $(call linux_version,$(file < $(KPATH)/include/generated/utsrelease.h))

 DRIVER            := 1
 MMAKE_USE_KBUILD   := 1
@@ -43,6 +46,6 @@ greater_version = $(shell printf "$(1)\n$(2)\n" | sort --version-sort | tail -1)
 # This approach does not work on Linux < 5.4, so in order not to break old
 # kernels, do it only with Linux >= 6.0.
 KBUILD_LIB_MAKE_TRG ?=
-ifeq ($(call greater_version,$(LINUX_VERSION_3),6.0), $(LINUX_VERSION_3))
+ifeq ($(call greater_version,$(LINUX_VERSION_N),6.0), $(LINUX_VERSION_N))
 KBUILD_LIB_MAKE_TRG += $(lib_obj_path)
 endif
```

It allows 2+ numbers in the version string and removes annoying escape characters with `-r`.

I also think it would be nice to add some unit-test coverage for this tentative `linux_version()` whenever this Makefile is included, and perhaps for the `greater_version()` a few lines below too, but I haven't figured out how to express it better in Makefile.

What's your opinion?

okt-sergeyn: Sorry for long response.
Thank you for such a good testing and your proposal. It seems much more reliable.
However, `<<<` will require setting `SHELL = /bin/bash`

> add some unit-test coverage

Do you mean something like this?
```
ifneq ($(call linux_version,#define UTS_RELEASE "6.5.0-14-generic"), 6.5.0)
$(error "linux_version error")
endif
```

ivatet-amd (approved):

ivatet-amd:
Hello @okt-sergeyn, could you re-test the latest `master` please? Shortly after merging, we discovered a couple of issues in older RHELs with GNU Make 3.82 and 4.2, fixed in https://github.com/Xilinx-CNS/onload/commit/1c8a11c94bbb21c96b8206a1183206f8bccaa356.

okt-sergeyn:
> Hello @okt-sergeyn, could you re-test the latest `master` please? Shortly after merging, we discovered a couple of issues in older RHELs with GNU Make 3.82 and 4.2, fixed in [1c8a11c](https://github.com/Xilinx-CNS/onload/commit/1c8a11c94bbb21c96b8206a1183206f8bccaa356).

Yes, it still works as supposed. Thanks for a quick fix.
Sorry for adding extra job for you :)

PR #205, opened 2024-02-15
kcompat: fix file refcounting issue in oo_fd_replace_file()

okt-sergeyn:
Recent linux-6.7 compat patch replaced lookup_fd_rcu() with lookup_fdget_rcu(). They are not identical though, because the latter takes reference to the file.

The patch adds extra fput() and fixes the wrapper to be compatible with older kernels.

Fixes: e34a911603af ("driver: compat for lookup_fd_rcu()")

### DIFF
--- a/src/driver/linux_onload/onload_kernel_compat.h
+++ b/src/driver/linux_onload/onload_kernel_compat.h
@@ -115,18 +115,5 @@ static inline int efrm_unregister_netdevice_notifier(struct notifier_block *b)
 #define unregister_netdevice_notifier efrm_unregister_netdevice_notifier
 #endif
 
-/* Linux-6.7 introduces lookup_fdget_rcu().
- * 5.11 >= linux < 6.7 - lookup_fd_rcu().
- * linux < 5.11 - fcheck(). */
-#ifndef EFRM_HAS_LOOKUP_FDGET_RCU
-static inline struct file *lookup_fdget_rcu(unsigned int fd)
-{
-#ifdef EFRM_HAS_LOOKUP_FD_RCU
-  return lookup_fd_rcu(fd);
-#else
-  return fcheck(fd);
-#endif
-}
-#endif
 
 #endif /* __ONLOAD_KERNEL_COMPAT_H__ */
--- a/src/driver/linux_onload/ossock_calls.c
+++ b/src/driver/linux_onload/ossock_calls.c
@@ -81,6 +81,8 @@ static int
 oo_fd_replace_file(struct file* old_filp, struct file* new_filp,
                    int old_fd, int* new_fd_p)
 {
+  struct file* tmp_filp = NULL;
+
   task_lock(current);
   if( atomic_read(&current->files->count) != 1 ) {
     /* This is a multithreaded application, and someone can be already
@@ -111,16 +113,18 @@ oo_fd_replace_file(struct file* old_filp, struct file* new_filp,
     return 0;
   }
 
-  rcu_read_lock();
-  if( lookup_fdget_rcu(old_fd) != old_filp ) {
-    rcu_read_unlock();
+  tmp_filp = fget(old_fd);
+  if( tmp_filp != old_filp ) {
+    if( tmp_filp )
+      fput(tmp_filp);
     task_unlock(current);
     return -EINVAL;
   }
 
+  fput(tmp_filp);
+
   get_file(new_filp);
   rcu_assign_pointer(files_fdtable(current->files)->fd[old_fd], new_filp);
-  rcu_read_unlock();
   task_unlock(current);
 
   /* No synchronize_rcu() is needed here.  See do_dup2() for an example,
--- a/src/driver/linux_resource/kernel_compat.sh
+++ b/src/driver/linux_resource/kernel_compat.sh
@@ -124,8 +124,6 @@ EFRM_HAS_REMAP_VMALLOC_RANGE_PARTIAL	export	remap_vmalloc_range_partial	include/
 
 EFRM_HAS_KTIME_GET_REAL_SECONDS	export	ktime_get_real_seconds	include/linux/timekeeping.h	kernel/time/timekeeping.c
 EFRM_FILE_HAS_F_EP	member	struct_file	f_ep	include/linux/fs.h
-EFRM_HAS_LOOKUP_FD_RCU	symbol	lookup_fd_rcu	include/linux/fdtable.h
-EFRM_HAS_LOOKUP_FDGET_RCU	symbol	lookup_fdget_rcu	include/linux/fdtable.h
 
 EFRM_HAS_FLUSH_DELAYED_FPUT	export	flush_delayed_fput	include/linux/file.h	fs/file_table.c
 


src/driver/linux_onload/onload_kernel_compat.h
```diff
@@ -121,11 +121,15 @@ static inline int efrm_unregister_netdevice_notifier(struct notifier_block *b)
...
+
 #ifdef EFRM_HAS_LOOKUP_FD_RCU
-  return lookup_fd_rcu(fd);
+  f = lookup_fd_rcu(fd);
 #else
-  return fcheck(fd);
+  f = fcheck(fd);
 #endif
+
+  return f ? get_file(f) : NULL;
```
ol-alexandra, line 132: I do not understand.  In the case of "new" kernel, are we taking 2 refcounts, one in `lookup_fd_rcu`, and another in `get_file`?

ivatet-amd: How about if we implement a function something like `bool oo_check_file(int fd, struct file *filp)` that would either call `lookup_fd_rcu()` or `fcheck()` and then call `fput()` in the former case, so that the caller would not need to decrement the reference counter?

The `get_file()` seems excessive in our case.

src/driver/linux_onload/ossock_calls.c
```diff
@@ -112,12 +114,18 @@ oo_fd_replace_file(struct file* old_filp, struct file* new_filp,
   }
 
   rcu_read_lock();
-  if( lookup_fdget_rcu(old_fd) != old_filp ) {
+  tmp_filp = lookup_fdget_rcu(old_fd);
+  if( tmp_filp != old_filp ) {
+    if( tmp_filp )
+      fput(tmp_filp);
     rcu_read_unlock();
     spin_unlock(&current->files->file_lock);
```
ivatet-amd, line 122: Since we're here...

Do you know where we call the corresponding `spin_lock(&current->files->file_lock)`? Maybe it should call `task_unlock(current);` instead?

src/driver/linux_onload/onload_kernel_compat.h
okt-sergeyn: > I do not understand. In the case of "new" kernel, are we taking 2 refcounts, one in `lookup_fd_rcu`, and another in `get_file`?

With "new" kernel we use kernel version of `lookup_fdget_rcu()`, which increments refcount.
With "old" kernels we use this wrapper, that does `lookup_fd_rcu()` + `get_file()`. 
`lookup_fd_rcu()` does not increment refcount.

okt-sergeyn: > How about if we implement a function something like `bool oo_check_file(int fd, struct file *filp)` that would either call `lookup_fd_rcu()` or `fcheck()`, compare the pointers and then call `fput()` in the former case, so that the caller would not need to decrement the reference counter?
> 
> The `get_file()` seems excessive in our case.

Good proposal. I'll rewrite it that way.

ol-alexandra: Got it, thank you!

src/driver/linux_onload/ossock_calls.c
ol-alexandra: `task_unlock` is `spin_unlock(&p->alloc_lock)`, which differ from `spin_unlock(&p->files->file_lock)`.

I guess you've found a real bug, because this error path was never tested.

ol-alexandra (approved):

src/driver/linux_onload/onload_kernel_compat.h
ivatet-amd: Ah, now I get it too!

There are three functions:
- `fcheck`
- `lookup_fd_rcu`
- `lookup_fdget_rcu`

The first two (used in Onload) do not bump the reference counter. The last one does, but Onload does not use it.

It means that as-is this PR does not fix anything because there is no resource leak, but still there is a possibility for us to handle all these three and call `fput` to cope with `lookup_fdget_rcu` which is the only available of these three in the latest kernel, right? In this case, the `oo_check_file()` suggestion still sounds adequate?

okt-sergeyn: Yes, in linux-6.7 we only have `lookup_fdget_rcu()`. After calling it we must do `fput()`.
In older kernels we have `lookup_fd_rcu`, in the oldest - `fcheck`.
I created a wrapper with the latest name and make it increment refcount also, to mimic new behavior. In such case the caller uses the latest interface regardless of kernel version.

We may implement `oo_check_file()` and deal with refcounting and compatibilty there.
For me both variants look reasonable, but I'd keep current solution.

ivatet-amd:
`oo_fd_replace_file()` attempts to replace a file structure for the given descriptor. If it is a multi-threaded application, it creates a new file descriptor. Otherwise, it tries to be as lightweight as possible and replaces the file object. However, since we are now bumping the reference counter, would it be simpler to call `fget()` instead of entering the RCU manually and bumping the reference counter for kcompat?

ol-alexandra:
> `oo_fd_replace_file()` attempts to replace a file structure for the given descriptor. If it is a multi-threaded application, it creates a new file descriptor. Otherwise, it tries to be as lightweight as possible and replaces the file object. However, since we are now bumping the reference counter, would it be simpler to call `fget()` instead of entering the RCU manually and bumping the reference counter for kcompat?

Good question!

src/driver/linux_onload/ossock_calls.c
ol-alexandra: See #206

src/driver/linux_onload/ossock_calls.c
```diff
@@ -112,12 +114,18 @@ oo_fd_replace_file(struct file* old_filp, struct file* new_filp,
   }
 
   rcu_read_lock();
-  if( lookup_fdget_rcu(old_fd) != old_filp ) {
+  tmp_filp = lookup_fdget_rcu(old_fd);
+  if( tmp_filp != old_filp ) {
+    if( tmp_filp )
+      fput(tmp_filp);
     rcu_read_unlock();
```
ol-alexandra, line 121: Why can't we use simple fget() here?

It is exactly what Ivan has asked.

ol-alexandra (changes requested):

src/driver/linux_onload/ossock_calls.c
ivatet-amd: Yup, we can remove `lookup_fd_rcu()` and the kcompat for it entirely.

okt-sergeyn: Thanks guys! That's done.
\+ rebase on top of master

ol-alexandra:
In the commit log: please mention that the new `lookup_fdget_rcu()` function in linux-6.7 takes a file reference in any case.

ivatet-amd:
Thanks @okt-sergeyn, I'll merge as soon as @ol-alexandra approves. 👍

ol-alexandra (approved):

Issue #207, opened 2024-03-01
linux-6.7 + huge pages: Failed to map packets!


okt-sergeyn:
Onload fails to create stack on Linux 6.7 with enabled huge pages:
```
$ onload nc -l 20002
oo:ncat[91964]: __ci_netif_pkt: mmap(0x0000000040000000) failed for pkt set 1 (0)
oo:ncat[91964]: Failed to map packets!
oo:ncat[91964]: FAIL at /home/nikitins/work/onload/src/lib/transport/ip/netif.c:1220
oo:ncat[91964]: ci_assert(ef_eplock_is_locked(&(ni)->state->lock))
from /home/nikitins/work/onload/src/lib/transport/ip/netif.c:1220

```
It works with `EF_USE_HUGE_PAGES=0` or in AF_XDP mode.

ivatet-amd:
This is caused by https://lore.kernel.org/lkml/20230907001832.GA63356@monkey/T/ where the page index became `PAGE_SIZE`-grained (and not hugepage-size grained).

ivatet-amd:
Must be fixed by https://github.com/Xilinx-CNS/onload/commit/d8201b2fc82bd56c73d89e3d16ed70c382fa15ec. 👍

PR #211, opened 2024-03-20
Fix Linux 6.8 build issues

okt-sergeyn:
This patch set fixes build on Linux 6.8.

Build works on my `Ubuntu 24.04 (Noble Numbat) 6.8.0-11-generic` setup.
I haven't tested it heavily, just checked tcp connection with netcat on AF_XDP.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,6 +146,11 @@ $(obj)/src/driver/linux_onload: $(obj)/src/lib/citools $(obj)/src/lib/ciul \
                                 $(obj)/src/lib/transport/ip \
                                 $(obj)/src/lib/cplane mkdirs
 
+# Linux-6.8 turns on -Wmissing-declarations and -Wmissing-prototypes globally.
+# Onload has too many places which trigger those warnings. So, just remove
+# the options to keep build working.
+KBUILD_CFLAGS := $(filter-out -Wmissing-prototypes -Wmissing-declarations, $(KBUILD_CFLAGS))
+
 else
 ################## Top-level makefile
 
--- a/src/lib/efhw/af_xdp.c
+++ b/src/lib/efhw/af_xdp.c
@@ -1416,13 +1416,58 @@ af_xdp_nic_buffer_table_clear(struct efhw_nic *nic,
  * Filtering
  *
  *--------------------------------------------------------------------*/
+
+static int
+af_xdp_ethtool_set_rxfh_context(struct efhw_nic *nic, const u32 *indir,
+                                const u8 *key, u8 hfunc, u32 *rss_context,
+                                u8 delete)
+{
+  struct net_device *dev = nic->net_dev;
+  const struct ethtool_ops *ops = dev->ethtool_ops;
+
+  EFHW_ASSERT(rss_context);
+
+#ifndef EFRM_HAVE_SET_RXFH_CONTEXT
+  /* linux >= 6.8 removes ethtool_ops::set_rxfh_context(). We use set_rxfh(). */
+
+  int rc;
+  struct ethtool_rxfh_param rxfh = {
+    .hfunc = hfunc,
+    .indir_size = nic->rss_indir_size,
+    .indir = (u32 *)indir,
+    .key_size = nic->rss_key_size,
+    .key = (u8 *)key,
+    .rss_context = *rss_context,
+    .rss_delete = delete,
+  };
+
+  if( !ops->set_rxfh ) {
+    EFHW_WARN("%s: %s does not support `set_rxfh` operation", __FUNCTION__,
+              dev->name);
+    return -EOPNOTSUPP;
+  }
+
+  rc = ops->set_rxfh(dev, &rxfh, NULL);
+  if( rc == 0 )
+    *rss_context = rxfh.rss_context;
+
+  return rc;
+#else
+  if( !ops->set_rxfh_context ) {
+    EFHW_WARN("%s: %s does not support `set_rxfh_context` operation",
+              __FUNCTION__, dev->name);
+    return -EOPNOTSUPP;
+  }
+  return ops->set_rxfh_context(dev, indir, key, hfunc, rss_context, delete);
+#endif
+}
+
 static int
 af_xdp_rss_alloc(struct efhw_nic *nic, const u32 *indir, const u8 *key,
 		 u32 efhw_rss_mode, int num_qs, u32 *rss_context_out)
 {
 	struct net_device *dev = nic->net_dev;
 	int rc = 0;
-	const struct ethtool_ops *ops;
 
 	EFHW_ASSERT(efhw_rss_mode == EFHW_RSS_MODE_DEFAULT);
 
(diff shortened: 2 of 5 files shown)


ivatet-amd:
@okt-sergeyn Believe it or not, I have an equivalent patch series internally in flight to address the kernel compat issues, which I want to merge today or tomorrow!

The summary:
1. A minute before you filed this PR, I was looking at the (long) list of warnings when built with `MMAKE_LIBERAL=1` and was sitting on the fence about how to best handle it. On my list, I see `-Wmissing-prototypes` but not `-Wmissing-declarations`. It means we should probably take your approach because if I fixed all `-Wmissing-prototypes`, it would not fix your case.
2. `strscpy()`: I'm adding a new kernel compat because the fix also goes into the older Enterprise Onload branch, where we potentially support older kernels without that function, so I want to keep it.
3. AF_XDP: I was not considering it, and willing to take your patch.
4. I also have a fix for https://github.com/Xilinx-CNS/onload/issues/207.

Let me apply (1) and (3) from your PR and (2) and (4) from mine if that sounds good.

My box is Ubuntu 23.10 with Linux 6.8.1 from the mainline kernel PPA.

okt-sergeyn:
> @okt-sergeyn Believe it or not, I have an equivalent patch series internally in flight to address the kernel compat issues, which I want to merge today or tomorrow!
> 
> The summary:
> 
>     1. A minute before you filed this PR, I was looking at the (long) list of warnings when built with `MMAKE_LIBERAL=1` and was sitting on the fence about how to best handle it. On my list, I see `-Wmissing-prototypes` but not `-Wmissing-declarations`. It means we should probably take your approach because if I fixed all `-Wmissing-prototypes`, it would not fix your case.
> 
>     2. `strscpy()`: I'm adding a new kernel compat because the fix also goes into the older Enterprise Onload branch, where we potentially support older kernels without that function, so I want to keep it.
> 
>     3. AF_XDP: I was not considering it, and willing to take your patch.
> 
>     4. I also have a fix for [linux-6.7 + huge pages: Failed to map packets! #207](https://github.com/Xilinx-CNS/onload/issues/207).
> 
> 
> Let me apply (1) and (3) from your PR and (2) and (4) from mine if that sounds good.
> 
> My box is Ubuntu 23.10 with Linux 6.8.1 from the mainline kernel PPA.

Thanks for quick response!
Please proceed with your plan. I'm good with that.

src/lib/efhw/af_xdp.c
```diff
@@ -1416,13 +1416,59 @@ af_xdp_nic_buffer_table_clear(struct efhw_nic *nic,
...
+  struct ethtool_rxfh_param rxfh = {
+    .hfunc = hfunc,
+    .indir_size = nic->rss_indir_size,
+    .key_size = nic->rss_key_size,
+    .rss_context = *rss_context,
+    .rss_delete = delete,
+  };
+
+  memcpy(rxfh.indir, indir, sizeof(indir[0]) * nic->rss_indir_size);
+  memcpy(rxfh.key, key, nic->rss_key_size);
```
tcrawley-xilinx, line 1443: Wouldn't `rxfh.indir` and `rxfh.key` both be 0 here? So this would end up writing to null.

I'd think it would be fine just setting `.indir = indir` and `.key = key` in the initialiser above.

tcrawley-xilinx (commented):
(only a review of the AF_XDP stuff)

Makefile
```diff
@@ -146,6 +146,11 @@ $(obj)/src/driver/linux_onload: $(obj)/src/lib/citools $(obj)/src/lib/ciul \
                                 $(obj)/src/lib/transport/ip \
                                 $(obj)/src/lib/cplane mkdirs
 
+# Linux-6.8 turns on -Wmissing-declarations and -Wmissing-prototypes globally.
+# Onload has too many places which trigger those warnings. So, just remove
+# the options to keep build working.
+KBUILD_CFLAGS := $(filter-out -Wmissing-prototypes -Wmissing-declarations, $(KBUILD_CFLAGS))
```
ivatet-amd, line 152: I'm considering these (similar) changes for the new and old build systems:
```diff
diff --git a/Makefile b/Makefile
index fd610837b2..4e92360569 100644
--- a/Makefile
+++ b/Makefile
@@ -213,6 +213,9 @@ ifneq ($(MMAKE_LIBERAL),1)
 ONLOAD_CFLAGS += -Werror
 endif

+# FIXME Address these in the source code.
+ONLOAD_CFLAGS += -Wno-missing-prototypes -Wno-missing-declarations
+
 ONLOAD_MAKEFLAGS ?=

 ifeq ($(HAVE_SFC),1)
diff --git a/mk/linux_kbuild.mk b/mk/linux_kbuild.mk
index 4d617478a7..c842decfc0 100644
--- a/mk/linux_kbuild.mk
+++ b/mk/linux_kbuild.mk
@@ -30,6 +30,9 @@ endif
 ifndef MMAKE_LIBERAL
 EXTRA_CFLAGS += -Werror
 endif # MMAKE_LIBERAL
+#
+# FIXME Address these in the source code.
+EXTRA_CFLAGS += -Wno-missing-prototypes -Wno-missing-declarations

 ifdef W_NO_STRING_TRUNCATION
 EXTRA_CFLAGS += -Wno-stringop-truncation
```
As they seem less intrusive. Please let me know if it doesn't work for you.

src/lib/efhw/af_xdp.c
okt-sergeyn: > Wouldn't `rxfh.indir` and `rxfh.key` both be 0 here? So this would end up writing to null.

omg, you're right. Thanks for noticing that.

> I'd think it would be fine just setting `.indir = indir` and `.key = key` in the initialiser above.

`indir` and `key` are passed as const, so compiler doesn't allow to do that: `error: initialization discards ‘const’ qualifier from pointer target type`.
Probably we can just discard const by doing like this `.indir = (u32*)indir`. I will recheck it.

Makefile
okt-sergeyn: Looks good!

ivatet-amd: Awesome. 👍 I'm happy to merge once the AF_XDP stuff is ready.

src/lib/efhw/af_xdp.c
okt-sergeyn: I've looked through some kernel and sfc netdriver code. It seems we can cast these pointers to non-const safely.
For instance, netdriver developers do the same in their compat code: https://github.com/Xilinx-CNS/onload/blob/6be7c86d813082d4b1bb48961f79e9123e9b572f/src/driver/linux_net/drivers/net/ethernet/sfc/ethtool_common.c#L2105

okt-sergeyn:
I've updated AF_XDP part according to @tcrawley-xilinx notes.
@ivatet-amd, please let me know if I should update "missing-prototypes" commit as well, or rebase the branch upon top of master.

ivatet-amd (approved):
Thank you @okt-sergeyn, no worries, I'll rebase/drop/update myself. 👍

Let's wait @tcrawley-xilinx approval and merge it.

tcrawley-xilinx (approved):
AF_XDP stuff looks good!

ivatet-amd:
Got this ready to merge in my checkout, running automated smoke-testing.

ivatet-amd:
Merged, thank you!

Issue #212, opened 2024-03-25
aarch64 build is broken with newdriver build


okt-sergeyn:
Onload master 13f04401
Building modules on raspberry pi with 5.15.83 kernel:
```
HAVE_EF10CT=0 HAVE_EFCT=0 HAVE_SFC=0 mmakebuildtree --driver
HAVE_EF10CT=0 HAVE_EFCT=0 HAVE_SFC=0 make -C build/aarch64_linux-5.15.83-v8/
```
Errors:
```
MODPOST /home/pi/onload/build/aarch64_linux-5.15.83-v8/Module.symvers                                                                                                                                         
ERROR: modpost: "__aarch64_ldset4_acq_rel" [/home/pi/onload/build/aarch64_linux-5.15.83-v8/src/driver/linux_resource/sfc_resource.ko] undefined!                                                                
ERROR: modpost: "__aarch64_cas4_acq_rel" [/home/pi/onload/build/aarch64_linux-5.15.83-v8/src/driver/linux_resource/sfc_resource.ko] undefined!                                                                  
ERROR: modpost: "__aarch64_ldclr4_acq_rel" [/home/pi/onload/build/aarch64_linux-5.15.83-v8/src/driver/linux_onload/onload.ko] undefined!                                                                        
ERROR: modpost: "__aarch64_ldadd4_acq_rel" [/home/pi/onload/build/aarch64_linux-5.15.83-v8/src/driver/linux_onload/onload.ko] undefined!                                                                        
ERROR: modpost: "__aarch64_ldset4_acq_rel" [/home/pi/onload/build/aarch64_linux-5.15.83-v8/src/driver/linux_onload/onload.ko] undefined!                                                                        
ERROR: modpost: "__aarch64_swp8_acq" [/home/pi/onload/build/aarch64_linux-5.15.83-v8/src/driver/linux_onload/onload.ko] undefined!                                                                              
ERROR: modpost: "__aarch64_cas8_acq_rel" [/home/pi/onload/build/aarch64_linux-5.15.83-v8/src/driver/linux_onload/onload.ko] undefined!                                                                          
ERROR: modpost: "__aarch64_cas4_acq_rel" [/home/pi/onload/build/aarch64_linux-5.15.83-v8/src/driver/linux_onload/onload.ko] undefined!                                                                          
[...]

PR #215, opened 2024-04-18
issue #212: fix aarch64 build with newdriver

okt-sergeyn:
Pass -mno-outline-atomics for gcc>=10, like it would work with old-style driver build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,6 +176,8 @@ $(eval $(patsubst CONFIG_%,export CONFIG_%$(_LF), \
        $(shell echo -e '$(_GET_NET_CONFIG_OPTS)' | \
                MFLAGS= MAKEFLAGS= make -C src/driver/linux_net/drivers/net/ethernet/sfc -r --no-print-directory -f - print_vars)))
 
+gcc_maj_ver := $(shell ./scripts/mmaketool --gcc_major_version)
+
 # CFLAGS
 ONLOAD_CFLAGS += -I$$(obj) -I$$(obj)/src -I$$(src) -I$$(src)/src -I$$(src)/src/include \
                  -D__ci_driver__ "-DTRANSPORT_CONFIG_OPT_HDR=<$(TRANSPORT_CONFIG_OPT_HDR)>"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -176,6 +176,9 @@ $(eval $(patsubst CONFIG_%,export CONFIG_%$(_LF), \
        $(shell echo -e '$(_GET_NET_CONFIG_OPTS)' | \
                MFLAGS= MAKEFLAGS= make -C src/driver/linux_net/drivers/net/ethernet/sfc -r --no-print-directory -f - print_vars)))
 
+gcc_maj_ver := $(shell ./scripts/mmaketool --gcc_major_version)
+arch := $(shell ./scripts/mmaketool --processor)
```
rhughes-xilinx, line 180: Isn't this duplicative of `KARCH`?

ivatet-amd (approved):

Makefile
```diff
@@ -238,6 +240,14 @@ else
   ONLOAD_CFLAGS += -DCI_HAVE_SFC=0
 endif
 
+test-ge = $(shell test $1 -ge $2 && echo 1)
+
+ifneq ($(call test-ge, $(gcc_maj_ver), 10), )
+ifeq ($(KARCH), aarch64)
```
rhughes-xilinx, line 246: We could make the build microscopically faster by turning these conditions around (and making `gcc_maj_ver` delay-evaluated, but I measured it as only ~30ms.

rhughes-xilinx (approved):

Makefile
ivatet-amd: It actually seems more logical to check for the arch first and then GCC specifically to apply the arch-specific fix.

Is there anything else to do except this?
```make
gcc_maj_ver = $(shell ./scripts/mmaketool --gcc_major_version)
...
ifeq ($(KARCH), aarch64)
ifneq ($(call test-ge, $(call gcc_maj_ver), 10), )
```

okt-sergeyn: I thought there might be other common (non arch specific) things to be set in case of gcc>=10. That's why I put it first.

ivatet-amd: Fair enough, I won't pursue this suggestion as it is a marginal difference.

Thanks for fixing the ARM64 built time.

PR #218, opened 2024-04-25
Unset CC which breaks onload DKMS build

elcabaye:
When `CC` is set, it appears to switch into cross-compile mode which breaks the DKMS build.

### DIFF
--- a/scripts/onload_misc/dkms.conf
+++ b/scripts/onload_misc/dkms.conf
@@ -7,9 +7,9 @@ PACKAGE_VERSION=
 # We quote make to avoid a misfeature of DKMS which adds a definition
 # of KERNELRELEASE, which should only be defined by kbuild.
 if which onload_uninstall >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


ivatet-amd:
Thank you for the PR!

Please help us understand the root cause of the suggested patch:

1. Are there multiple compilers installed in the host?
2. What is currently passed as `CC` so that the patch wants to unset it?
3. Where does the wrong `CC` come from?

These are the first questions I can think of, but could you please provide more relevant information, such as build/DKMS logs, if that would also help?

ivatet-amd:
We came across this issue in RHEL9 and DKMS 3.0.13 so that it brings more clarity, and I can answer my questions above:

1. No, there is a single toolchain in the affected host.
2. `CC=gcc` and `KERNEL_CC=gcc`, and it triggers the cross-compilation behaviour in Onload.
3. DKMS 3.0.13 adds these two environment variables in addition to DKMS 3.0.11 variables.

We believe that, ideally, Onload should handle `CC=` differently. However, we anticipate that some users rely on Onload to trigger cross-compilation, and we do not want to disrupt their workflow at the moment.

It makes me more comfortable to merge this PR. If it breaks somebody's workflow, this can help us justify the effort to review the Onload build scripts.

PR #220, opened 2024-05-15
Improve splice() support

okt-sergeyn:
Support splice() from Onload socket to unaccelerated pipe.
Refactor existing compat code.

The PR addresses the `usecases/splice` issue. @okt-yurijp please check this out.


ivatet-amd:
I want to take a look within the next few days, but not immediately, because we're preparing for the upcoming release.

src/lib/efthrm/tcp_helper_linux.c
```diff
@@ -60,15 +60,45 @@
...
-    return base_handler(iocb->ki_filp, iter_iov(v), v->nr_segs);        \
+    return base_handler(iocb->ki_filp, iter_iov(v), v->nr_segs, CI_ADDR_SPC_CURRENT); \
   } while (0);
 
 #endif /* EFRM_HAVE_ITER_UBUF */
 
+static ssize_t rw_iter_bvec_handler(ssize_t(*base_handler)(struct file*,
+                                                           const struct iovec*,
+                                                           unsigned long,
+                                                           ci_addr_spc_t),
```
ol-alexandra, line 114: Please define a type for this `base_handler`.  With an explanatory comment.

ol-alexandra (approved):

ol-alexandra (approved):

src/lib/efthrm/tcp_helper_linux.c
```diff
@@ -60,15 +68,83 @@
...
+
+  iov.iov_base = kzalloc(len, GFP_KERNEL);
+  iov.iov_len = len;
+
+  rc = base_handler(iocb->ki_filp, &iov, 1, CI_ADDR_SPC_KERNEL);
+  /* This handler is called only in splice_read() case, i.e. from
+   * socket to pipe. We don't have to care about another direction,
+   * copy_to_iter() is enough. */
+  if( rc > 0 )
+    rc = copy_to_iter(iov.iov_base, iov.iov_len, v);
```
ivatet-amd, line 95: Could you clarify why we pass `iov.iov_len` and not `rc` to `copy_to_iter()`? AFAIU, `iov.iov_len == iov_iter_count()` is the capacity, but `rc` is the number of bytes actually "retrieved" from the base handler such as `linux_tcp_helper_fop_read_iov_pipe()`, so it seems we don't want to exceed `rc`.

ivatet-amd (approved):
Thank you, it looks good to me. I'll merge the PR tomorrow morning. 👍

Could you review one outstanding question in `rw_iter_pipe_handler()`?

src/lib/efthrm/tcp_helper_linux.c
okt-sergeyn: Yes, you're right! `rc` must be there. Thanks.
I fixed that.

ivatet-amd: Excellent, thank you!

ivatet-amd:
Merged, but with an incorrect merge message (copied and pasted it from somewhere without editing the commit message).

ivatet-amd:
I also made a follow-up kernel compat fix https://github.com/Xilinx-CNS/onload/commit/06eae317dd000824e085b47aa69cc1917c558228 for older kernels like 3.10 in RHEL7. (Please ping me if there is something you would do differently.)

okt-sergeyn:
Another fix for older kernels - https://github.com/Xilinx-CNS/onload/pull/222

PR #225, opened 2024-05-31
epoll3: fix socket cleanup after duplicating epoll fd

okt-sergeyn:
After closing a dup()'ed epoll fd all the sockets which were added to that fd get invalid epoll reference in citp_fdinfo. In epoll3 case it leads to a refcounting issue - when the socket is closed we do not release reference to netif taken in citp_epoll_set_home_stack().

Keep list of all opened epoll fds associated with this epoll set and set valid epoll_fd to all sockets referenced to the epoll fd which is being closed.

### DIFF
--- a/src/lib/transport/unix/epoll_fd.c
+++ b/src/lib/transport/unix/epoll_fd.c
@@ -451,6 +451,22 @@ static void citp_epoll_cleanup_home_sock_list(struct citp_epoll_fd* ep,
     oo_wqlock_unlock(&ep->dead_stack_lock, NULL);
   }
 }
+
+static void citp_epoll_update_eitems_epoll_fd(ci_dllist *list, int epoll_fd,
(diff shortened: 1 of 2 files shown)


ivatet-amd:
Planning to take a look and test drive early next week. 👍

ivatet-amd:
Tested before and after applying the fix with `EF_UL_EPOLL=3`:

1. Create a socket.
2. Create an epoll instance.
3. Duplicate an epoll instance.
4. Add socket (1) to the duplicated epoll instance (3).
5. Close duplicated epoll instance (3).
6. Close socket (1).
7. Close the originating epoll instance (2).

Before the fix, the netif reference count was 2. After the fix, it became 1, as expected. (For AMD readers, I uploaded the testing notes to ON-15837.)

About to merge, thanks. 👍

ivatet-amd (approved):

PR #227, opened 2024-06-12
lib/ip: handle socket buffer maximum values in Linux way

okt-sergeyn:
According to man SO_RCVBUF and SO_SNDBUF socket options are limited by /proc/sys/net/core/[rw]mem_max values.


ol-alexandra (approved):
I believe that the commit log should mention the TCP word.

ivatet-amd (approved):
Thanks, looks good. The kernel also clips the size quietly: https://elixir.bootlin.com/linux/v6.9.3/source/net/core/sock.c#L1227.

PR #230, opened 2024-06-24
Fix fdtable allocation on new Debian

### LINKED ISSUE #229: citp_fdtable_ctor: failed to allocate fdtable
Debian testing/sid:
```
oo:ncat[1582]: citp_fdtable_ctor: failed to allocate fdtable (0x3ffffff8)
oo:ncat[1582]: FAIL at /home/sasha/work/level5/onload/src/lib/transport/unix/startup.c:685
oo:ncat[1582]: EtherFabric transport library: failed to initialise (9)
oo:ncat[1582]: hostname=elrond pid=1582
/home/sasha/work/level5/onload/build/gnu_x86_64_debt/lib/transport/unix/libcitransport0.so(+0x109f4d)[0x7f235a08df4d]
/home/sasha/work/level5/onload/build/gnu_x86_64_debt/lib/transport/unix/libcitransport0.so(+0xb4b1)[0x7f2359f8f4b1]
/lib64/ld-linux-x86-64.so.2(+0x4e35)[0x7f235a0f2e35]
/lib64/ld-linux-x86-64.so.2(+0x4f64)[0x7f235a0f2f64]
/lib64/ld-linux-x86-64.so.2(+0x1a840)[0x7f235a108840]
Aborted (core dumped)
```
The issue is, Onload uses `RLIMIT_NOFILE:rlim_max` value as fdtable size.  With the recent changes in systemd the soft limit is the same as previously (1024, select-compatible) but hard limit is increased to 0x3ffffff8.  System can't allocate that amount of memory for fdtable.

okt-sergeyn:
The first commit fixes #229 
The second one fixes initialization of `citp_log_level`.

### DIFF
--- a/src/include/ci/internal/opts_citp_def.h
+++ b/src/include/ci/internal/opts_citp_def.h
@@ -403,6 +403,11 @@ CI_CFG_OPT("EF_FDTABLE_SIZE", fdtable_size, ci_uint32,
 "(help ulimit, man 2 setrlimit) are bound by this value.",
            , , 0, MIN, MAX, count)
 
+CI_CFG_OPT("EF_FDS_MAX", fds_max, ci_uint32,
+"The value which is set as RLIMIT_NOFILE hard limit if the system value "
+"exceeds it. EF_FDTABLE_SIZE overrides it.",
+           , , 1048576, MIN, MAX, count)
+
 #define CI_UL_LOG_E     0x1            /* errors */
 #define CI_UL_LOG_U     0x2            /* unexpected */
 #define CI_UL_LOG_S     0x4            /* setup */
--- a/src/lib/transport/unix/fdtable.c
+++ b/src/lib/transport/unix/fdtable.c
@@ -500,6 +500,15 @@ int citp_fdtable_ctor()
   citp_fdtable.size = 4096;
 
   if( getrlimit(RLIMIT_NOFILE, &rlim) == 0 ) {
+    /* Handle the case when rlim_max is too huge for malloc(). */
+    if( rlim.rlim_max > CITP_OPTS.fds_max ) {
+      Log_S(ci_log("Limit max number of opened files to EF_FDS_MAX=%u value.",
+                   CITP_OPTS.fds_max));
+      rlim.rlim_max = CITP_OPTS.fds_max;
+      if( rlim.rlim_cur > rlim.rlim_max )
+        rlim.rlim_cur = rlim.rlim_max;
+      CI_TRY(ci_sys_setrlimit(RLIMIT_NOFILE, &rlim));
+    }
     citp_fdtable.size = rlim.rlim_max;
     if( CITP_OPTS.fdtable_size != 0 &&
         CITP_OPTS.fdtable_size != rlim.rlim_max ) {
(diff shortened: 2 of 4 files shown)


ol-alexandra (approved):

ivatet-amd:
Thank you for filing a PR. 👍

I cannot schedule it at the moment, but we are keeping it on our radar.

Presumably, in the absence of a fix in Debian, a workaround for #229 is passing `EF_FDTABLE_SIZE` to Onload. This PR makes this workaround go away. Is my understanding correct?

What obstacles do you anticipate for the dynamic approach, where the FD tables increase in runtime on demand dynamically (say, unless `EF_FDTABLE_SIZE` is set, which could prevent the dynamic behaviour to avoid jitter)?

okt-sergeyn:
> Presumably, in the absence of a fix in Debian, a workaround for #229 is passing `EF_FDTABLE_SIZE` to Onload. This PR makes this workaround go away. Is my understanding correct?

Yes, absolutely right.

> What obstacles do you anticipate for the dynamic approach, where the FD tables increase in runtime on demand dynamically (say, unless `EF_FDTABLE_SIZE` is set, which could prevent the dynamic behaviour to avoid jitter)?

It'd be a major modification I guess. Maybe @ol-alexandra has some thoughts.

ol-alexandra:
> > What obstacles do you anticipate for the dynamic approach, where the FD tables increase in runtime on demand dynamically (say, unless `EF_FDTABLE_SIZE` is set, which could prevent the dynamic behaviour to avoid jitter)?
> 
> It'd be a major modification I guess. Maybe @ol-alexandra has some thoughts.

I agree that it would be a major modification with unpleasent experience for some users.  For example some users run a multithreaded application with `EF_FDS_MT_SAFE=1`.  If one thread re-allocates fdtable, and other thread assumes lockless access to fdtable, it would result in a surprise.  You can say that such users must set `EF_FDTABLE_SIZE`, but it is not an easy task to pass this message to all the users.

You can implement "dynamic approach" which does not include `realloc()` call (and does not change the user address), but again, it is a major rework of fdtable.

New Debian will probably release next summer; you have some months to decide.  At the same time, as this change comes from systemd, the same issue may arise with other distros (Fedora and so on).

okt-sergeyn:
Hello! Friendly asking.
Are there any updates on this issue?

sianj-xilinx:
> Are there any updates on this issue?

Hi Sergey. We have been thinking about this, but haven't yet come to a conclusion. The problem you point out definitely needs addressing, but I've got a few concerns about this. The main issue is that this will change the rlimit by default, which has the potential to silently break (unusual for onload) applications. The benefit of doing this way is obviously that things Just Work for most applications. The new `EF_FDS_MAX` is also very similar to the existing `EF_FDTABLE_SIZE`. We're considering what the best approach is, with alternatives including this or a more explicit configuration, for example bailing out with a helpful error message if we fail to allocate the fdtable, in which case the user can either configure the app or system limit, or use the existing `EF_FDTABLE_SIZE`. In the medium term we'd like to move to a dynamic approach, but in practice this will probably become a problem before we get that work done, so we'll need an interim solution.

okt-sergeyn:
Hi Siân! I got your point. Thanks for detailed answer.

PR #232, opened 2024-07-09
lib/ip: fix {SND,RCV}BUFFORCE socket options limits

okt-sergeyn:
Fix computing the limits for *_BUFFORCE socket options.
Take into account missing PKTS_PER_SET constant.

### DIFF
--- a/src/include/ci/internal/ip.h
+++ b/src/include/ci/internal/ip.h
@@ -3117,6 +3117,11 @@ ci_inline int ci_netif_pkt_release_check_keep(ci_netif* ni, ci_ip_pkt_fmt* pkt)
   }
 }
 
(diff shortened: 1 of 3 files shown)


ol-alexandra (approved):

ivatet-amd (approved):
That is scary because it is virtually impossible to notice the miscalculation until you use the socket options.

I believe the components breakdown is the following:
- `CI_CFG_PKT_BUF_SIZE` is the packet buffer size, in bytes.
- `PKTS_PER_SET` is the buffer set size, in packet buffers.
- `pkt_sets_max()` is the maximum number of buffer sets.

Thus, `ci_netif_pkt_sets_max_size()` returns the maximum size in bytes across all buffer sets. Previously, the equivalent expression returned a value that does not have meaning.

PR #238, opened 2024-08-07
eflatency: Optionally echo the packet in the pong reply and support VLAN tags

osresearch:
This patch adds the option of having the `pong` node copy the contents of the `ping` message into the reply, which adds a little more realism to the `eflatency` test since it requires the receiver to read the contents of the message, not just receive the notice that a message has arrived.

Additionally, it also adds the option for 802.1q VLAN tagging for `eflatency` tests that traverse switches, making it possible to benchmark those switches as well.

It also cleans up a bit of the logic by removing some magic sizes by using `sizeof()` on various ethernet headers.

### DIFF
--- a/src/tests/ef_vi/eflatency.c
+++ b/src/tests/ef_vi/eflatency.c
@@ -39,6 +39,7 @@ static inline void rx_wait_with_ts(struct eflatency_vi*);
 
 static int              cfg_verbose = 0;
 static int              cfg_validate = 0;
+static int              cfg_vlan = 0;
 static int              cfg_iter = 100000;
 static int              cfg_warmups = 10000;
 static int              cfg_payload_len = DEFAULT_PAYLOAD_SIZE;
@@ -65,7 +66,8 @@ static enum ef_vi_flags cfg_vi_flags = 0;
 #define BUF_SIZE        2048
 #define MAX_UDP_PAYLEN	(1500 - sizeof(ci_ip4_hdr) - sizeof(ci_udp_hdr))
 /* Protocol header length: Ethernet + IP + UDP. */
-#define HEADER_SIZE     (14 + 20 + 8)
+#define ETH_HEADER_SIZE (cfg_vlan ? sizeof(ci_ethhdr_vlan_t) : sizeof(ci_ethhdr_t))
+#define HEADER_SIZE     (ETH_HEADER_SIZE + sizeof(ci_ip4_hdr) + sizeof(ci_udp_hdr))
 
 
 struct pkt_buf {
@@ -103,8 +105,7 @@ const uint16_t port_he = 8080;
 
 static void checksum_udp_pkt(void * pkt_buf)
 {
-  ci_ether_hdr * const eth = (ci_ether_hdr*) pkt_buf;
-  ci_ip4_hdr * const ip4 = (void*) ((uint8_t*) eth + 14);
+  ci_ip4_hdr * const ip4 = (void*) ((uint8_t*) pkt_buf + ETH_HEADER_SIZE);
   ci_udp_hdr * const udp = (void*) (ip4 + 1);
   struct iovec iov = {
     .iov_base = udp + 1,
@@ -118,8 +119,9 @@ static void checksum_udp_pkt(void * pkt_buf)
 
 static void init_udp_pkt(void* pkt_buf, int paylen)
 {
-  ci_ether_hdr * const eth = (ci_ether_hdr*) pkt_buf;
-  ci_ip4_hdr * const ip4 = (void*) ((uint8_t*) eth + 14);
+  ci_ethhdr_t * const eth = pkt_buf;
+  ci_ethhdr_vlan_t * const eth_vlan = pkt_buf;
+  ci_ip4_hdr * const ip4 = (void*) ((uint8_t*) pkt_buf + ETH_HEADER_SIZE);
   ci_udp_hdr * const udp = (void*) (ip4 + 1);
   const size_t ip_len = sizeof(*ip4) + sizeof(*udp) + paylen;
 
@@ -129,7 +131,15 @@ static void init_udp_pkt(void* pkt_buf, int paylen)
 
   memcpy(eth->ether_dhost, remote_mac, sizeof(remote_mac));
   ef_vi_get_mac(&rx_vi.vi, driver_handle, eth->ether_shost);
-  eth->ether_type = htons(0x0800);
+
+  if( !cfg_vlan ) {
+    eth->ether_type = CI_ETHERTYPE_IP;
+  } else {
+    eth_vlan->ether_type = CI_ETHERTYPE_IP;
+    eth_vlan->ether_vtype = CI_ETHERTYPE_8021Q;
+    eth_vlan->ether_vtag = htons(cfg_vlan);
+  }
+
   ci_ip4_hdr_init(ip4, CI_NO_OPTS, ip_len, 0, IPPROTO_UDP, htonl(laddr_he),
                   htonl(raddr_he), 0);
   ci_udp_hdr_init(udp, ip4, htons(port_he), htons(port_he), udp + 1, paylen, 0);
@@ -733,6 +743,7 @@ static CI_NORETURN usage(const char* fmt, ...)
   fprintf(stderr, "  -n <iterations>     - set number of iterations\n");
   fprintf(stderr, "  -s <message-size>   - set udp payload size. Accepts Python slices\n");
   fprintf(stderr, "  -w <iterations>     - set number of warmup iterations\n");
+  fprintf(stderr, "  -V <vlan-tag>       - add vlan tag to ethernet header\n");
   fprintf(stderr, "  -c <cut-through>    - CTPIO cut-through threshold\n");
   fprintf(stderr, "  -p                  - CTPIO no-poison mode\n");
   fprintf(stderr, "  -m <modes>          - allow mode of the set: [c]tpio, \n");
@@ -778,7 +789,7 @@ int main(int argc, char* argv[])
     p = (unsigned int)__v;                                   \
   } while( 0 );
 
-  while( (c = getopt (argc, argv, "n:s:w:c:pm:t:o:vr")) != -1 )
+  while( (c = getopt (argc, argv, "n:s:w:c:pm:t:o:V:vr")) != -1 )
     switch( c ) {
     case 'n':
       OPT_INT(optarg, cfg_iter);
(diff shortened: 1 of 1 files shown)


src/tests/ef_vi/eflatency.c
```diff
@@ -214,24 +222,40 @@ generic_ping(struct eflatency_vi* rx_vi, struct eflatency_vi* tx_vi,
...
     uint64_t start = ci_frc64_get();
     tx_send(tx_vi);
     if( do_rx_post )
       rx_post(&rx_vi->vi);
     rx_wait(rx_vi);
+
+    if (cfg_validate && cfg_payload_len > 0)
+    {
+      const uint8_t rx_pattern = rx_vi->rx_pkt[HEADER_SIZE];
+      if (pattern != rx_pattern)
```
jfeather-amd, line 264: I wonder if when validating it's worth validating the whole packet using `memcp` for example, or if for more detail about the first octet that's different a custom loop would suffice.

src/tests/ef_vi/eflatency.c
```diff
@@ -130,6 +132,13 @@ static void init_udp_pkt(void* pkt_buf, int paylen)
   memcpy(eth->ether_dhost, remote_mac, sizeof(remote_mac));
   ef_vi_get_mac(&rx_vi.vi, driver_handle, eth->ether_shost);
   eth->ether_type = htons(0x0800);
+
+  if (cfg_vlan) {
+    eth_vlan->ether_type = htons(0x0800);
+    eth_vlan->ether_vtype = htons(0x8100);
```
jfeather-amd, line 138: You could use `CI_ETHERTYPE_8021Q` here instead of `htons(0x8100)` if you like.

src/tests/ef_vi/eflatency.c
```diff
@@ -130,6 +132,13 @@ static void init_udp_pkt(void* pkt_buf, int paylen)
   memcpy(eth->ether_dhost, remote_mac, sizeof(remote_mac));
   ef_vi_get_mac(&rx_vi.vi, driver_handle, eth->ether_shost);
   eth->ether_type = htons(0x0800);
```
jfeather-amd, line 134: I think I would prefer if this was moved into an `else` under the `if( cfg_vlan )` check. I understand that this is inoffensive because you correctly write this data region, but I would prefer not to incorrectly write to `eth_vlan->ether_vtype` here if we are using a VLAN.

src/tests/ef_vi/eflatency.c
```diff
@@ -130,6 +132,13 @@ static void init_udp_pkt(void* pkt_buf, int paylen)
...
+
+  if (cfg_vlan) {
```
jfeather-amd, line 136: nit: style
```diff
-  if (cfg_vlan) {
+  if( cfg_vlan ) {
```

jfeather-amd (commented):
Thanks for the contribution, @osresearch!

This review just covers things I found by inspection, and I plan on doing some testing in the coming days, but overall I quite like these changes! I do have one concern in regards to performance which is perhaps a non-issue, I will confer with other members of the team in regards to this concern. There are also quite a few style nit-picks, please feel free to disregard these if you want - I can apply these changes in the merging process if you would rather focus on the code itself.

src/tests/ef_vi/eflatency.c
```diff
@@ -99,34 +101,40 @@ const uint32_t raddr_he = 0xac010203;  /* 172.1.2.3 */
 const uint16_t port_he = 8080;
 
 
+static void checksum_udp_pkt(void * pkt_buf)
+{
+  ci_ether_hdr * const eth = (ci_ether_hdr*) pkt_buf;
+  ci_ip4_hdr * const ip4 = (void*) ((uintptr_t) eth + 14);
```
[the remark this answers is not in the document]
osresearch: `uintptr_t` implies to me that the (integer) result will be offset by a number of bytes, although `uint8_t*` or `char*` would work too.  Also note that the magic number `14` (copied from the original code) gets replaced with `ETH_HEADER_SIZE` in the commit that adds VLAN support.

src/tests/ef_vi/eflatency.c
```diff
@@ -214,24 +222,40 @@ generic_ping(struct eflatency_vi* rx_vi, struct eflatency_vi* tx_vi,
...
-   generic_desc_check(tx_vi, 0);
+    generic_desc_check(tx_vi, 0);
   }
 
   gettimeofday(&start, NULL);
 
-  for( i = 0; i < cfg_iter; ++i ) {
+  for( i = 0; i < cfg_iter; ++i, pattern++ ) {
+    memset(&tx_pkt[HEADER_SIZE], pattern, cfg_payload_len);
+    checksum_udp_pkt(tx_pkt);
```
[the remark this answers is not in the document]
osresearch: The `memset()` and `checksum_udp_pkt()` calls are outside of the timing loop for the `ping` process, so it shouldn't affect the measurement results.  It might add a minimal amount of overhead to the send, although since there are also `printf()` and other calls that the process does as part of reporting the results, I'm not sure if you could even tell...
On the receive side (either `pong` or the `ping` process receiving the reply), it seems to add about 150ns per side on my test machine.

src/tests/ef_vi/eflatency.c
```diff
@@ -475,8 +512,12 @@ generic_desc_check(struct eflatency_vi* vi, int wait)
 
       switch( EF_EVENT_TYPE(*ev) ) {
       case EF_EVENT_TYPE_RX:
+        // sfc driver
+        vi->rx_pkt = pkt_bufs[EF_EVENT_RX_RQ_ID(*ev)]->dma_buf;
         return;
       case EF_EVENT_TYPE_RX_REF:
+        // efct driver
+        vi->rx_pkt = efct_vi_rxpkt_get(&vi->vi, ev->rx_ref.pkt_id);
```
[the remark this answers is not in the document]
osresearch: That was mostly for me as I figured out how to update the code to find the X3 receive address...

Is it safe to get the pointer like this and then call `handle_rx_ref()` or does that needs to be deferred to after the contents of the packet have been used?

src/tests/ef_vi/eflatency.c
osresearch: The first version of the patch (as you noted above) only set the first byte... I can add a loop to check the rest of them.

osresearch:
Thanks for the feedback on the patch.  I'll make the style corrections and push an updated version.

src/tests/ef_vi/eflatency.c
```diff
@@ -214,24 +222,39 @@ generic_ping(struct eflatency_vi* rx_vi, struct eflatency_vi* tx_vi,
...
     tx_send(tx_vi);
     if( do_rx_post )
       rx_post(&rx_vi->vi);
     rx_wait(rx_vi);
+
+    if( cfg_validate && cfg_payload_len > 0 ) {
+      const uint8_t rx_pattern = rx_vi->rx_pkt[HEADER_SIZE];
+      if( pattern != rx_pattern )
+        fprintf(stderr, "expected pong %02x got %02x\n", pattern, rx_pattern);
+    }
```
jfeather-amd, line 266: After you pointed out the memory operations were outside of the timing loop, I spotted that this bit of code isn't. I would be interested to see if your 150ns time increase changes if this it moved after `uint64_t stop = ci_frc64_get();`

jfeather-amd (commented):
Thanks for addressing my comments so quickly! I'm still looking at some tests for this, and have kicked off a test run to go overnight.

src/tests/ef_vi/eflatency.c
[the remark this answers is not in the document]
jfeather-amd: I will preface this comment by saying that I understand how incredibly pedantic this note is and indeed, as you have mentioned, any of these will really work in practice. Also, apologies if the below is all stuff that you know - I only learnt the difference between some of these things a few months ago and thought the difference was quite fascinating!

I think my hesitance is mainly because the C standard doesn't really define arithmetic on `uintptr_t`, as far as my understanding goes. The standard itself only defines the type such that a valid `void*`, cast to `uintptr_t`, then cast back into a `void*` will necessarily be the same as the original pointer.

In the case of pointer arithmetic, I think it's best to define the exact type/unit you're traversing in memory. At the time of the second commit (the diff that the original comment is on), the magic `14` is used as the length of an ethernet frame (without VLAN tagging). In that situation, we want to traverse exactly 14 octets, so we should use a `uint8_t*`.

However, because this is later changed to drop the magic `14` and replace it with a conditional `sizeof` operation, I would argue that `uint8_t*` is no longer the appropriate unit. The unit of the data returned by `sizeof` is defined to be a byte, which means we should be using a `char*`.

All that being said, I notice you have changed this to use `uint8_t*`, and I imagine we would be in a far stickier situation than eflatency having dodgy headers if compiled on a machine where 1 byte isn't 8 bits.

jfeather-amd:
Hi @osresearch, sorry for the delay in getting back to you on this! I just finished looking into performance testing this patch and found that there seems to be a significant enough regression that I am hesitant to merge this in its current state. I would like to think for a while longer about how to progress this PR, as I do think this would be a nice change to have! Some options to consider are:
- Making a similar change to a different ef_vi test app
- Locking this behind a build option
- Duplicating eflatency to have eflatency_memops (for example) which incorporates this change

Although I haven't thought for long enough to decide which one of these would be most appropriate.

osresearch:
Thanks for doing the performance testing on the patch, @jfeather-amd .  Can you describe where the slowdowns seem to be?  In the non-vlan, non-echo, non-validating case (the default), my latency deltas were in the noise on the X2 and X3 cards, so I'm very curious about your methodology so that I can replicate the results for my future testing.

osresearch:
I've re-run tests on the X3 cards with better isolation and pinning the `eflatency` task to a single CPU; the results show no change in the `min`, `50%`, `95%` and `99%` numbers, although there is an unexpected increase in the `mean` of about 50ns.  This is caused by the unconditional  `memset()` and `checksum_udp_pkt()` on the send side, although these occur outside of the `ci_frc64_get()` timing loop and which I had assumed would not affect the timing. Adding `if(cfg_validating)...` around the packet rewriting removes this effect.

However, this performance regression appears to be an issue with the way `mean` is computed -- it is the *total* time for all packets (delta between the two `gettimeofday()` calls), not the mean of the measured times (`rdtsc` ticks).  I wonder if the `mean` should be computed as the average of the actual times instead.  It is unexpected to me that the first column of results doesn't match the data used for the other columns.

ivatet-amd:
Hello @osresearch, sorry for not replying sooner. We are busy!

The purposes why we use `eflatency` include performance comparisons between various OpenOnload versions on different hardware configurations. If we change the methodology behind `eflatency` reports, we won't be able to compare them between the released OpenOnload versions without the change and newer ones with the change. ("Comparing apples and oranges".) It might not be a technical problem, but it is still significant work to update the ecosystem to accommodate this change gracefully without confusing the users (e.g. "Why did the performance unexpectedly become better/worse?"), and we are not motivated to do it now.

Would you like to contact our customer support team (and sales team) so we can better understand your motivation and usage cases and figure out what we can do about them? You would need to email [support-nic@amd.com](mailto:support-nic@amd.com).

PR #239, opened 2024-08-16
lib/ip: calculate SO_RCVBUFFORCE limit more accurately

okt-sergeyn:
Do not allow socket receive buffer size to exceed maximum number of RX packets.

### DIFF
--- a/src/include/ci/internal/ip.h
+++ b/src/include/ci/internal/ip.h
@@ -3117,9 +3117,23 @@ ci_inline int ci_netif_pkt_release_check_keep(ci_netif* ni, ci_ip_pkt_fmt* pkt)
   }
 }
 
-ci_inline ci_uint32 ci_netif_pkt_sets_max_size(ci_netif* ni)
+/* Functions below are used to compute maximum socket buffer size limit when
+ * setting it via RCVBUFFORCE/SNDBUFFORCE socket options. The aim is to not
+ * allow user to set a buffer big enough to let a single socket grab all the
+ * packets available for the stack.
+ * Additionally, the returned value must be halved before comparing with user's
+ * request to behave as Linux. See usage in ci_set_sol_socket() and
+ * ci_udp_setsockopt_lk(). */
+ci_inline ci_uint32 ci_netif_max_rx_packets_size_per_socket(const ci_netif* ni)
 {
-  return pkt_sets_max(ni) * PKTS_PER_SET * CI_CFG_PKT_BUF_SIZE;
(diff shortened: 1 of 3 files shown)


ol-alexandra:
This patch tells: it we have one socket, then it should not consume all RX packets from the stack.  But this patch allows to consume all RX packets using 2 sockets.  I.e. the patch is an improvement comparing to #232.  But I'm not sure that the number "2" is well-chosen.

ol-alexandra (approved):

okt-sergeyn:
> This patch tells: it we have one socket, then it should not consume all RX packets from the stack. But this patch allows to consume all RX packets using 2 sockets. I.e. the patch is an improvement comparing to #232. But I'm not sure that the number "2" is well-chosen.

Thanks for the note.
Looking at the further code
```c
    s->so.rcvbuf = CI_MAX(oo_adjust_SO_XBUF(v), (int) NI_OPTS(netif).tcp_rcvbuf_min);
```
```c
/* src/lib/transport/ip/ip_internal.h */
#define oo_adjust_SO_XBUF(v)  ((v) * 2)
```
The value is doubled afterwards. So we still can consume all RX packets with single socket.
I believe this "/2" is needed to meet socket API.

ivatet-amd:
I assumed `max_rx_packets` and `pkt_sets_max` are related, i.e. one is a function of the other. Thus it looks like further clarification of https://github.com/Xilinx-CNS/onload/pull/232/ and OK.

What about `SO_SNDBUFFORCE` and `max_tx_packets`? Would this patch apply the same way to TX?

ivatet-amd (approved):

ol-alexandra:
> The value is doubled afterwards. So we still can consume all RX packets with single socket. I believe this "/2" is needed to meet socket API.

Could you please change your "/2" to something like "/5"?  I believe it is completely wrong that one socket can consume all the RX packets leaving no packets for RX ring.

ol-alexandra (changes requested):
Please change "/2" -> "/5"

ivatet-amd:
@ol-alexandra Shall we worry about the existing users if we reduce the limit?

I suppose we're waiting for @okt-sergeyn to come back to the PR. In addition to `/2` and `SO_SNDBUFFORCE` comments, may I ask to describe a couple of use cases demonstrating the changes in action? Is it a no-op as-is?

okt-sergeyn:
Use the same approach for TX.
Replace "/2" with bigger numbers.
@ol-alexandra please confirm that I got your idea right.

ol-alexandra (approved):

PR #240, opened 2024-08-28
eflatency: compute the mean from the timings array

osresearch:
This patch changes how the mean column in eflatency is computed to use the sum of the per-message timings, instead of the overall time that the test used.  That way things that are outside of the timing loop do not affect the computation of the mean and the mean value reflects the min/50/90/95/max values in the other columns.

For most users this will result in a slight decrease of their mean times, although ideally it should be very close for the default case.

### DIFF
--- a/src/tests/ef_vi/eflatency.c
+++ b/src/tests/ef_vi/eflatency.c
@@ -151,13 +151,13 @@ static void output_results(struct timeval start, struct timeval end)
 {
   unsigned freq = 0;
   double div;
+  int i;
   int usec = (end.tv_sec - start.tv_sec) * 1000000;
   usec += end.tv_usec - start.tv_usec;
 
   ci_get_cpu_khz(&freq);
   div = freq / 1e3;
   if( cfg_save_file ) {
-    int i;
     char* subst = strstr(cfg_save_file, "$s");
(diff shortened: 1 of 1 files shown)


ivatet-amd:
Thank you @osresearch.

Although I understand the motivation, I don't think we can apply this patch because it might confuse the users and give a wrong impression of the performance improvement since the `mean` would decrease.

Meanwhile, I'll let @jfeather-amd follow up at https://github.com/Xilinx-CNS/onload/pull/238.

osresearch:
I see that a similar change has been merged in the v9.0.0 branch in yaml output mode https://github.com/Xilinx-CNS/onload/commit/dd61f46b29be9b5323c812575b900e35f3d017a3 so this can be closed.

pemberso-xilinx:
> I see that a similar change has been merged in the v9.0.0 branch in yaml output mode [dd61f46](https://github.com/Xilinx-CNS/onload/commit/dd61f46b29be9b5323c812575b900e35f3d017a3) so this can be closed.

Thanks for the initial suggestion @osresearch .  If you want to raise any other PR building on dd61f46b29be9b5323c812575b900e35f3d017a3 , we're happy to take a look.

Issue #236, opened 2024-07-29
CentOS 9 Linux 6.10: No rule to make target 'src/lib/citools/buddy.o', needed by src/lib/citools/lib.a'


okt-sergeyn:
Driver build is broken on CentOS 9 with Linux 6.10:
```
$ KVER=6.10.1-1.el9.elrepo.x86_64 HAVE_SFC=0 make
make[1]: Entering directory '/usr/src/kernels/6.10.1-1.el9.elrepo.x86_64'
(/home/nikitins/work/onload/src/driver/linux_resource/kernel_compat.sh -k /usr/src/kernels/6.10.1-1.el9.elrepo.x86_64 -q > /home/nikitins/work/onload/build/x86_64_linux-6.10.1-1.el9.elrepo.x86_64/src/driver/linux_resource/autocompat.h) || (rm -f /home/nikitins/work/onload/build/x86_64_linux-6.10.1-1.el9.elrepo.x86_64/src/driver/linux_resource/autocompat.h && false)
make[4]: *** No rule to make target '/home/nikitins/work/onload/build/x86_64_linux-6.10.1-1.el9.elrepo.x86_64/src/lib/citools/buddy.o', needed by '/home/nikitins/work/onload/build/x86_64_linux-6.10.1-1.el9.elrepo.x86_64/src/lib/citools/lib.a'.  Stop.
make[3]: *** [scripts/Makefile.build:485: /home/nikitins/work/onload/build/x86_64_linux-6.10.1-1.el9.elrepo.x86_64/src/lib/citools] Error 2
make[2]: *** [/usr/src/kernels/6.10.1-1.el9.elrepo.x86_64/Makefile:1934: /home/nikitins/work/onload/build/x86_64_linux-6.10.1-1.el9.elrepo.x86_64] Error 2
make[1]: *** [Makefile:240: __sub-make] Error 2
make[1]: Leaving directory '/usr/src/kernels/6.10.1-1.el9.elrepo.x86_64'
make: *** [Makefile:276: modules] Error 2
```

okt-sergeyn:
fixed in #242

PR #242, opened 2024-09-16
Fix root makefile to work with linux 6.10

okt-sergeyn:
Linux 6.10 has broken Onload kernel modules build by replacing $(src) with $(obj) in its makefiles.
See Linux commit 9a0ebe5011f49e932bb0a2cea2034fd65e6e567e

This patch implements a possible approach - create symlinks to Onload source files in build directory.

---

Possible fix for #236
Checked on CentOS 9 `6.10.5-1.el9.elrepo.x86_64` and Ubuntu 22.04 `6.8.0-40-generic` by running `make`/ `mmakebuildtree --driver && make`.

Note: there are other build issues, related to linux-6.10, which I'll address separately.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -282,7 +282,26 @@ clean_kernel:
 # Can't figure out a way to get modpost to look in the src directory. At least
 # the number of makefiles is much smaller than the number of source files
 $(OUTMAKEFILES): $(KBUILDTOP)/%: %
-	$(Q)mkdir -p $(@D)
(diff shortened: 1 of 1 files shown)


rhughes-xilinx:
One of my goals for this build system was not to have to do this kind of copying/linking (one reason being that it causes difficulties with deleted files), so it's disappointing to see it come back. I shouldn't be spending time on coming up with another intricate trick which avoids it, however. If somebody else has given it a good try and failed then I'll accept that conclusion

sianj-xilinx (approved):

PR #243, opened 2024-09-23
lib/ip: handle address space in udp tx/rx functions

okt-sergeyn:
Handle address space of a data pointer in udp tx/rx kernel-variant functions. It is needed for correct data copy to/from Onload sockets with kernel API. The usecase: calling splice to/from unaccelareted pipe from/to UDP socket.

@okt-yurijp please check it out


ol-alexandra (approved):

ivatet-amd (approved):
I suppose it's a follow-up since https://github.com/Xilinx-CNS/onload/pull/220.

PR #245, opened 2024-10-08
sfc_resource: make x86 syscall table search more reliable

okt-sergeyn:
Looking for 'e8' byte ("call" instruction) is error-prone. This patch adds an extra check for the obtained address to avoid possible kernel crash if it happens to be invalid.

### DIFF
--- a/src/driver/linux_resource/syscall_x86.c
+++ b/src/driver/linux_resource/syscall_x86.c
@@ -415,6 +415,14 @@ static bool find_syscall(void)
         result = (unsigned long)p1 + 5;
         result += p1[1] | (p1[2] << 8) | (p1[3] << 16) | (p1[4] << 24);
 
+        /* Check that the result address is sane. Do not allow next steps to
+         * read from the address if it does not point to a valid page within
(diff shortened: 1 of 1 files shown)


ivatet-amd (approved):
LGTM and I suggest @tcrawley-xilinx's review.

tcrawley-xilinx (approved):
Changes look good.
I'm curious, was this caused by a specific distro/kernel version? Or was it just found by inspection?

ivatet-amd:
I am also curious about why `e8` is error-prone, i.e. what does the false positive look like when it's not an address?

ol-alexandra (approved):

okt-sergeyn:
> Changes look good. I'm curious, was this caused by a specific distro/kernel version? Or was it just found by inspection?

I faced it accidentally on linux-6.8 debug kernel, compiled manually.

okt-sergeyn:
> I am also curious about why `e8` is error-prone, i.e. what does the false positive look like when it's not an address

Actually it looks like a real address. However, reading it causes error like this:
`BUG: unable to handle page fault for address: ffffffff24c00432`

PR #247, opened 2024-10-18
make: add capability to use old-style driver build

okt-sergeyn:
It can be used as a workaround in cases when new build system does not work for some reasons.


okt-sergeyn:
I use it as workaround for #236

ivatet-amd (approved):
LGTM, but I prefer @sianj-xilinx to approve too.

sianj-xilinx:
> LGTM, but I prefer @sianj-xilinx to approve too.

Let's merge #236 instead. That might not be optimal, but it's better than going back to having to maintain two build systems.

okt-sergeyn:
Closing it since #242 has been merged.

PR #250, opened 2024-10-30
linux-6.10: add compat for follow_pte()

okt-sergeyn:
The function prototype has been changed in linux 6.10. Support both variants to be able to run Onload on kernels 6.10 and 6.11.


sianj-xilinx (approved):
Thanks! I thought with all the variations we have here we should be covering all bases, but hadn't tested the specific 6.10/6.11 window.

okt-sergeyn:
Merged

PR #251, opened 2024-10-31
af_xdp: move rx timestamp read to support WODA

wdebruij:
Move AF_XDP rx timestamp read into vi->ops.receive_get_timestamp.

The original implementation added a branch in device independent get_rx_timestamp. This is not needed, and it breaks WODA, because record_rx_timestamp is skipped.

Fix this, and as a side-effect better isolate the AF_XDP from the device independent code.

WODA also needs the EF_VI_SYNC flags passed. XDP offers no way to query the device for these currently. Similar to ef10_mcdi_event, assume clock is always in sync.

With this change, src/tests/onload/wire_order succeeds.

### DIFF
--- a/src/lib/ciul/efxdp_vi.c
+++ b/src/lib/ciul/efxdp_vi.c
@@ -264,7 +264,25 @@ static void efxdp_ef_vi_receive_push(ef_vi* vi)
 static int efxdp_ef_vi_receive_get_timestamp(struct ef_vi* vi, const void* pkt,
                                              ef_precisetime* ts_out)
 {
-  return -EOPNOTSUPP;
+  const uint64_t ns_to_sec = 1000UL * 1000 * 1000;
+  const struct onload_xdp_rx_meta {
+#define ONLOAD_XDP_RX_META_TSTAMP 0x1
+    uint64_t flags;
+    uint64_t tstamp;
+  } *meta = pkt;
+
+  /* pkt points to start of packet data. meta precedes that */
+  meta--;
+  if (!(meta->flags & ONLOAD_XDP_RX_META_TSTAMP)) {
+    *ts_out = (ef_precisetime) { 0 };
+    return -ENODATA;
+  }
+
+  ts_out->tv_sec = meta->tstamp / ns_to_sec;
+  ts_out->tv_nsec = meta->tstamp % ns_to_sec;
+  ts_out->tv_nsec_frac = 0;
+  ts_out->tv_flags = 0;
+  return 0;
 }
 
 static void efxdp_ef_eventq_prime(ef_vi* vi)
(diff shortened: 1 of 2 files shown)


wdebruij:
Ping

jfeather-amd (approved):
Looks like a good change, thanks!

I've briefly tested this with some internal `woda_latency_{server,client}` apps and they seem to be happy to run over AF_XDP with this change, for internal reviewers, you can find the test log at `~jfeather/rhome/dev/public-pr-251/woda_latency_server.log`.

I would like to check what testing @ivatet-amd has done and why specifically he wanted @sianj-xilinx's review before merging this, but he is currently off, so I'll aim to come back to this PR in the coming weeks.

src/lib/ciul/efxdp_vi.c
```diff
@@ -264,7 +264,25 @@ static void efxdp_ef_vi_receive_push(ef_vi* vi)
...
+  meta--;
+  if (!(meta->flags & ONLOAD_XDP_RX_META_TSTAMP)) {
+    *ts_out = (ef_precisetime) { 0 };
+    return -ENODATA;
+  }
+
+  ts_out->tv_sec = meta->tstamp / ns_to_sec;
+  ts_out->tv_nsec = meta->tstamp % ns_to_sec;
+  ts_out->tv_nsec_frac = 0;
+  ts_out->tv_flags = EF_VI_SYNC_FLAG_CLOCK_SET | EF_VI_SYNC_FLAG_CLOCK_IN_SYNC;
```
abower-amd, line 284: Rather than fabricate traceability claims, would it be sufficient for WODA to claim only the `CLOCK_SET` flag and leave the `IN_SYNC` flag clear with `EF_TIMESTAMPING_REPORTING` kept at `0`?

wdebruij: I have updated the commit as in my previous comment.

wdebruij:
Please take another look if you have some time. The previous review comments (thanks!) are addressed.

src/lib/ciul/efxdp_vi.c
abower-amd: Thanks for that @wdebruij - my comment was not a blocking one, just a suggestion for consideration - I don't object to setting the flags that are needed for this use case if you find you need to re-introduce that later.

wdebruij: Thanks. No, on reviewing your comment I saw that I was setting the flags unnecessarily. So better to remove them (done).

jfeather-amd:
Apologies for the delay @wdebruij, I've merged this internally now and the change should make its way to this repo soon. Thank you for your patience!

wdebruij:
Thanks Josh!

PR #265, opened 2025-02-19
af_xdp: Allow running onload as non-root

zhuyifei1999:
There's a TODO saying how it fails when the process does not have CAP_NET_RAW. This is true, so to workaround that, swap to a temporary global-root cred from `prepare_kernel_cred(&init_task)`.

While in theory one could create a less-privileged cred via `prepare_creds()` and individually raise the required capability bits, just like in oo_cplane_empower_cap_net_raw, unfortunately xdp_umem_create -> xdp_umem_reg -> xdp_umem_account_pages requires CAP_IPC_LOCK in the init user ns, and escalating to init user ns while keeping capability bits sounds meaningless to me.

The netns of the process is backed by the nsproxy and not the cred so this should not affect netns. Though I'm not sure about the security implications of this, considering onload is able to run rootless in non-AF_XDP mode already.

### DIFF
--- a/src/lib/efhw/af_xdp.c
+++ b/src/lib/efhw/af_xdp.c
@@ -169,6 +169,7 @@ struct efhw_nic_af_xdp
   struct efhw_af_xdp_vi* vi;
   struct efhw_buddy_allocator vi_allocator;
   spinlock_t alloc_lock;
+  struct xdp_mmap_offsets mmap_offsets;
 };
 
 /*----------------------------------------------------------------------------
@@ -583,41 +584,39 @@ static int xdp_create_ring(struct socket* sock,
   return 0;
 }
 
-static int xdp_create_rings(struct socket* sock,
-                            struct efhw_page_map* page_map, void* kern_mem_base,
-                            long rxq_capacity, long txq_capacity,
-                            struct efab_af_xdp_offsets_rings* kern_offsets,
-                            struct efab_af_xdp_offsets_rings* user_offsets,
-                            struct ring_map* ring_mapping)
+static int xdp_get_mmap_offsets(struct xdp_mmap_offsets* mmap_offsets)
 {
-  int rc;
+  struct socket* sock;
   struct sys_call_area rw_area;
-  struct xdp_mmap_offsets* mmap_offsets;
+  struct xdp_mmap_offsets* mmap_offsets_user;
   int* optlen;
+  int rc;
 
-  EFHW_BUILD_ASSERT(EFAB_AF_XDP_DESC_BYTES == sizeof(struct xdp_desc));
+  rc = sock_create(AF_XDP, SOCK_RAW, 0, &sock);
+  if( rc < 0 )
+    return rc;
 
   /* We need a read-write area to call getsockopt().  We unmap it from UL
    * as soon as possible. */
   rc = sys_call_area_alloc(&rw_area);
   if( rc < 0 )
-    return rc;
+    goto out_release;
 
-  mmap_offsets = sys_call_area_ptr(&rw_area);
-  optlen = (void*)(mmap_offsets + 1);
-  *optlen = sizeof(*mmap_offsets);
+  mmap_offsets_user = sys_call_area_ptr(&rw_area);
+  optlen = (void*)(mmap_offsets_user + 1);
+  *optlen = sizeof(*mmap_offsets_user);
 
   /* For linux<=5.7 you can use kernel_getsockopt(),
    * but newer versions does not have this function, so we have all that
    * sys_call_area_*() calls. */
   rc = sock->ops->getsockopt(sock, SOL_XDP, XDP_MMAP_OFFSETS,
                              (void*)sys_call_area_user_addr(&rw_area,
-                                                            mmap_offsets),
+                                                            mmap_offsets_user),
                              (void*)sys_call_area_user_addr(&rw_area, optlen));
 
-  /* Security consideration: mmap_offsets is located in untrusted user
+  /* Security consideration: mmap_offsets_user is located in untrusted user
    * memory.  I.e. the process can overwrite all this data.
-   * However this is the process which can create an AF_XDP Onload stack,
+   * However this is the process which can load an XDP program,
    * so it runs with the root account, and it already can do
    * anything bad: reboot, execute arbitrary code, etc.
    *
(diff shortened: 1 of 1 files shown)


zhuyifei1999:
> Embed the offsets in the module with a couple offsetof() calls. 

It looks like `struct xdp_rxtx_ring` is in `net/xdp/xsk_queue.h` which is not exported. Grabbing the offset from UMH or BTF is complex, so rather than that, I'll grab it in `af_xdp_nic_init_hardware` which is trusted to be root.

ligallag-amd:
> AF_XDP rings gives non-root full access to a TX/RX ring (especially in zerocopy mode), allowing the user to masquarade as any sender, essentially CAP_NET_RAW. However, in sfc mode onload already provides CAP_NET_RAW on various code paths via oo_cplane_empower_cap_net_raw, so I don't think this is breaking an existing security boundary.

This seems fine to me.

> But somewhat more concerning is af_xdp.c's use of sys_call_area, especially for the XDP_MMAP_OFFSETS syscall. The syscall provides an memory offset, and if unchecked, gives path to a OOB memory access if userspace manages to race writing to that area before it is unmapped. I thought of two ways around this: 

I don't really understand the issue here - is the problem that an onloaded userspace app can make this syscall while privileges are breifly raised for the task. Why change in privileges cause this issue? Anyway, it sounds like you've found a solution to this and will update the PR accordingly.

zhuyifei1999:
> I don't really understand the issue here - is the problem that an onloaded userspace app can make this syscall while privileges are breifly raised for the task. Why change in privileges cause this issue? 

Looking at the code, specifically these two comments:
https://github.com/Xilinx-CNS/onload/blob/62861dbbe0d76a150926306b97c6d7bef7bb77c4/src/lib/efhw/af_xdp.c#L47-L58
https://github.com/Xilinx-CNS/onload/blob/62861dbbe0d76a150926306b97c6d7bef7bb77c4/src/lib/efhw/af_xdp.c#L618-L625

The assumption is that non-root isn't allowed to create af_xdp onload stacks, so userspace racing and overwriting mmap_offsets before it's unmapped isn't too bad, "because they are already root". However, if non-root is allowed to create af_xdp onload stacks, like what this patch does, it's much worse because now unprivileged users might break the kernel.

> Anyway, it sounds like you've found a solution to this and will update the PR accordingly.

Yep, just pushed. Sorry for the delay.

zhuyifei1999:
> Why does having the capability for CAP_IPC_LOCK mean we'd may as well raise the task capability further than strictly necessary?

The difference between the CAP_NET_RAW check and CAP_IPC_LOCK is the namespace in which the check is performed against. 

The CAP_NET_RAW checks against the user namespace owning the network namespace: https://github.com/torvalds/linux/blob/b4432656b36e5cc1d50a1f2dc15357543add530e/net/xdp/xsk.c#L1704-L1711

While the CAP_IPC_LOCK checks against the initial user namespace (initial user namespace being the namespace that's outside any containers, i.e. "global root", root of the machine rather than root of a container): https://github.com/torvalds/linux/blob/b4432656b36e5cc1d50a1f2dc15357543add530e/net/xdp/xdp_umem.c#L128-L133

The check here is not completely fatal, but rather constrains the umem size to the `RLIMIT_MEMLOCK` (max locked memory) rlimit. On a Rocky Linux 9 machine this was tested on, the max locked memory by default is hard limited to 8MiB:
```
$ ulimit -aH | grep locked
max locked memory           (kbytes, -l) 8192
```
.. which was insufficient.

So if onload wants to reliably run in af_xdp mode across distros, it might want to raise the rlimit, but to do that, it requires another capability, `CAP_SYS_RESOURCE`, in the initial user namespace, https://github.com/torvalds/linux/blob/b4432656b36e5cc1d50a1f2dc15357543add530e/kernel/sys.c#L1492-L1498

... which makes it redundant to just having `CAP_IPC_LOCK` to ignore the rlimit rather than having `CAP_SYS_RESOURCE` and raising the limit.

Now why checking against initial user namespace is so problematic? It's because of the interaction between capabilities and user namespaces. To quote the man page https://man7.org/linux/man-pages/man7/user_namespaces.7.html

> On the other hand, that process has no capabilities in the parent (in the case of clone(2)) or previous (in the case of unshare(2)) and setns(2)) user namespace, even if the new namespace is created or joined by the root user (i.e., a process with user ID 0 in the root namespace).

This means that if onload, running in a container with userns, were to set CAP_IPC_LOCK on itself, it would not work, because the initial user namespace is a parent and it has no capabilities in a parent user namespace. So if we'd want only strictly necessary capabilities, it would still mean, somehow jump to the initial user namespace, setting CAP_IPC_LOCK on itself. But then, to quote the man page again:

> If a process has a capability in a user namespace, then it has that capability in all child (and further removed descendant) namespaces as well.

It means it gains that capability across all namespaces (because the initial user ns is ancestor of everything else). And not to mention, there's no existing implementation of "jumping to initial user namespace with only certain capability bits set" in the kernel. I would want to empty the effective capability bits before applying bits like CAP_NET_RAW and CAP_IPC_LOCK. I'd need to implement "temporary switching user namespace but keep everything else", which... the closest thing in the kernel would be the setns syscall:

https://github.com/torvalds/linux/blob/b4432656b36e5cc1d50a1f2dc15357543add530e/kernel/nsproxy.c#L569-L581

... which, is not some exported kernel functions where I can say "ok ignore permissions and just do it" (normally you're not allowed return to an ancestor user namespace), but rather full of implementation details that I'm definitely uncomfortable messing around in, especially in an out-of-tree module.

Considering that the alternative is complex and would already give more permissions than strictly necessary (i.e. `CAP_NET_RAW` `CAP_IPC_LOCK` across all user namespaces), I just don't think it makes sense to pursue that direction instead of just using `prepare_kernel_cred` / `override_creds`. I could go and lower unneeded capability bits from the cred returned by `prepare_kernel_cred`, but tbh I don't see the value in that, given that nothing outside this function should ever be able to utilize that cred. It's functionally the same as "a root daemon created that fd and passed to this process with SCM_RIGHTS".

> Also, may I ask what sort of testing you've done?

This was tested a while back without the second patch so I don't fully remember, but IIRC we tested latency and WODA. We observed back packet losses to sshd (non-onload-accelerated), which I found to be related to flow steering not being fully functional and the data path of onload -> kernel was failing due to `inject_kernel_gid` restrictions. It was workarounded with `echo -1 > /sys/module/onload/parameters/inject_kernel_gid`. IIRC after that everything worked as expected.

With the second patch I only did basic functionality testing with netcat.

ligallag-amd:
Thanks for your detailed explanation YiFei - I've merged this internally and it should sync to our public repo.

PR #268, opened 2025-02-28
Fix linux-6.12 support

ol-alexandra:
Using Onload with Linux-6.12 results in crash with NULL pointer dereference.  It happens in many tests in Socket Tester, in nginx, in other cases.  In the git log I see that linux-6.12 support was added under ON-16108, but it definitely was not tested properly.

With these patches I see some basic tests passing on Linux-6.12.

### DIFF
--- a/src/lib/efthrm/tcp_helper_endpoint_move.c
+++ b/src/lib/efthrm/tcp_helper_endpoint_move.c
@@ -416,12 +416,10 @@ int efab_file_move_to_alien_stack(ci_private_t *priv, ci_netif *alien_ni,
   new_ep->file_ptr = priv->_filp;
 
   /* Copy F_SETOWN_EX, F_SETSIG to the new file */
-#ifdef F_SETOWN_EX
   rcu_read_lock();
   __f_setown(old_ep->alien_ref->_filp, efrm_file_f_owner(priv->_filp)->pid,
              efrm_file_f_owner(priv->_filp)->pid_type, 1);
   rcu_read_unlock();
-#endif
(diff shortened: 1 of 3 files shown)


src/lib/efthrm/tcp_helper_endpoint_move.c
```diff
@@ -416,12 +416,10 @@ int efab_file_move_to_alien_stack(ci_private_t *priv, ci_netif *alien_ni,
...
-  rcu_read_lock();
-  __f_setown(old_ep->alien_ref->_filp, efrm_file_f_owner(priv->_filp)->pid,
-             efrm_file_f_owner(priv->_filp)->pid_type, 1);
-  rcu_read_unlock();
-  efrm_file_f_owner(old_ep->alien_ref->_filp)->signum =
-                                      efrm_file_f_owner(priv->_filp)->signum;
+  rc = oo_copy_file_owner(old_ep->alien_ref->_filp, priv->_filp);
+  if( rc != 0 )
+    goto fail4;
+
```
ligallag-amd, line 422: I wonder if another failure case is needed here? Given that we have just alloc'd with `onload_alloc_file`, do we need something like: `fput(priv->_filp`? This is what is done in `onloadfs.c:oo_create_fd`. I'm pretty shaky in this area so I'm not sure!

ligallag-amd (commented):
Looks good. I just have one question

src/lib/efthrm/tcp_helper_endpoint_move.c
ol-alexandra: Thank you.  Fixed.

ligallag-amd (approved):

PR #274, opened 2025-05-09
TPH mode in ef10 isn't ever enabled.

lukester1975:
This is due to the compatibility version of `pcie_tph_get_cpu_st()` always being used (`EFRM_HAVE_LINUX_TPH_H` isn't defined before including `ci/efhw/tph.h`).

Here's a one line change to include `ci/driver/kernel_compat.h` in the same fashion as in `ef10ct.c`, though maybe it's better for `ci/efhw/tph.h` to include it?

Testing: enabling steering on X2522 no longer reports EINVAL.

Regards

### DIFF
--- a/src/include/ci/efhw/tph.h
+++ b/src/include/ci/efhw/tph.h
@@ -1,6 +1,8 @@
 /* SPDX-License-Identifier: GPL-2.0 OR BSD-2-Clause */
 /* X-SPDX-Copyright-Text: Copyright (C) 2024, Advanced Micro Devices, Inc. */
 
+#include <ci/driver/kernel_compat.h>
+
 #ifdef EFRM_HAVE_LINUX_TPH_H
 #include <linux/pci-tph.h>
 #else


jfeather-amd:
Thanks for spotting this! I was perhaps a little lax in my testing as I assumed the `-EINVAL` was due to lack of hardware support.

> Here's a one line change to include ci/driver/kernel_compat.h in the same fashion as in ef10ct.c, though maybe it's better for ci/efhw/tph.h to include it?

I'm inclined to agree - I would like it if `ci/efhw/tph.h` included it so that we can't run into this issue again. It does at least seem like there's [some precedent](https://github.com/search?q=repo%3AXilinx-CNS%2Fonload+path%3A*.h+%22%3Cci%2Fdriver%2Fkernel_compat.h%3E%22&type=code) for including the kernel compat file in another header file.

lukester1975:
OK, wasn't sure about removing the `kernel_compat.h` include from `ef10ct.c` so this is now just the additional include in `ci/efhw/tph.h`.

(If removed, it does still get included by some other route but I wouldn't like to figure out whether it's correct, so I'll leave it there...)

jfeather-amd (approved):
> OK, wasn't sure about removing the kernel_compat.h include from ef10ct.c so this is now just the additional include in ci/efhw/tph.h.

I probably would have done the same - it's safer to leave it as it is!

PR #275, opened 2025-05-09
Fix compilation with GCC 15, which defaults to C23.

lukester1975:
Nothing else to add!

Thanks

### DIFF
--- a/src/include/ci/efhw/common_ci2linux.h
+++ b/src/include/ci/efhw/common_ci2linux.h
@@ -11,7 +11,7 @@
 #endif
 #include <stdbool.h>
 
-#if ! (defined bool) && ! (defined __cplusplus)
+#if ! (defined bool) && ! (defined __cplusplus) && (__STDC_VERSION__ < 202311L)
 #undef false
 #undef true
 typedef enum {


src/include/ci/efhw/common_ci2linux.h
```diff
@@ -11,7 +11,7 @@
 #endif
 #include <stdbool.h>
 
-#if ! (defined bool) && ! (defined __cplusplus)
+#if ! (defined bool) && ! (defined __cplusplus) && (__STDC_VERSION__ < 202311L)
```
jfeather-amd, line 14: Could we change this to be an inclusive check of the final version that supports stdbool instead (i.e., C17 with `201710L`)? Testing with gcc-13.2.0 which admittedly only partially supports C23 (using `--std=c2x`) this check is bypassed because (presumably) a placeholder value was used for `__STDC_VERSION__`:
```
$ cat test.c
#include <stdio.h>

int main() {
  printf("__STDC_VERSION__ = %ld\n", __STDC_VERSION__);
  return 0;
}

$ gcc test.c -o test --std=c2x
$ ./test
__STDC_VERSION__ = 202000
```

Updating the clause as below worked for me:
```suggestion
#if ! (defined bool) && ! (defined __cplusplus) && (__STDC_VERSION__ <= 201710L)
```

Would it also be possible to add a brief comment explaining that C23 deprecated this header (and give context that the magic number represents C17) please?

jfeather-amd (commented):
Looks like a good change, thanks! I'd like it if you slightly modified the check to support compilers with partial C23 support (e.g., gcc-13.2.0). Also, this seems like something that could be beneficial on older versions of onload, so would it be possible to rebase this change onto the `v8_1` branch? Typically we target the oldest applicable branch for changes that seem more like fixes, rather than new features, and merge forwards (so the change will still make its way to the `v9_0` branch).

src/include/ci/efhw/common_ci2linux.h
lukester1975: Sure, done.

jfeather-amd (approved):
Thanks for updating this! I have realised in hindsight that my suggestion might break compilers with partial support that report a newer `__STDC_VERSION__` but haven't promoted booleans to first class citizens yet (though I haven't checked if such a version exists).

PR #279, opened 2025-05-27
Do not allow hw filter stealing in no-5tuple support scenarios

mina:
Currently we're running into the following behavior in AF_XDP based deployment (which is 3-tuple only support):

- Onload stack A starts and listens on port X.
- Onload stack A exits; its cleanup of hw filters is delayed.
- Onload stack B starts and listens on the same port X.
- Onload stack B sees the 3-tuple hw filter inserted by stack A, which has not yet been cleaned.
- In the current code, stack B will reuse the hw filter of stack A, but that 3-tuple hw filter points to the wrong stack, and the wrong RX queue.
- This causes all the sockets created by stack B to become unaccelerated, because they land on the rxq that belonged to stack A, not the rxq that belongs to stack B.

The root cause of this issue is that oof_socket_add_wild expects oof_local_port_addr_fixup_wild to fix hw filters that point to different stacks, but that will not actually happen if
(!oof_socket_can_share_hw_filter && fm->fm_hwports_no5tuple). This creates the undesirable behavior that all sockets in this new stack being unaccelerated, even when the user passed EF_NO_FAIL=0 (!)

Fix this by returning an error in this edge case. This means that a new onload stack will not reuse a hw filter that doesn't belong to it. The user can restart the workload when the old onload stack has finished cleanup of hw filters.

Note that this patch will change the behavior for all fm->fm_hwports_no5tuple deployments. This seems beneficial as all those deployments should see the same bug, but other options are possible:
- We could add a module parameter that configures this behavior.
- We may be able to only disable this behavior for AF_XDP deployments.

tested by running these commands with onload in a tight loop:

onload tcp_rr -F 1000
onload udp_rr -F 1000

tcp_rr and udp_rr reuse the same ports by default. In all cases, onload will start successfully if there is no existing hw filter left over from the previous onload invocation, and will fail with -EEXIST if there is a left over filter from the previous stack.

Change-Id: I6e196ac55cc81efbc537206cdc6e20b2068a7e3d

### DIFF
--- a/src/lib/efthrm/oof_filters.c
+++ b/src/lib/efthrm/oof_filters.c
@@ -2548,14 +2548,23 @@ oof_socket_add_wild(struct oof_manager* fm, struct oof_socket* skf,
   else if( ! oof_socket_can_share_hw_filter(skf, &lpa->lpa_filter) ) {
     /* H/w filter already exists but points to a different stack.  This is
      * fixed if necessary in oof_local_port_addr_fixup_wild().
+     *
+     * oof_local_port_addr_fixup_wild will not fix the hw filters if
+     * (!oof_socket_can_share_hw_filter && fm->fm_hwports_no5tuple). This will
+     * result in this stack reusing a hw filter that points to a different
+     * stack, resulting in unaccelerated sockets in this stack. Return an error
+     * here to prevent this.
      */
-    OO_DEBUG_IPF(other_skf = oof_wild_socket(lp, lpa, skf->af_space);
-                 if( other_skf != NULL )
-                   ci_log(FSK_FMT "STEAL "IPX_TRIPLE_FMT" from "SK_FMT,
-                          FSK_PRI_ARGS(skf),
-                          IPX_TRIPLE_ARGS(lp->lp_protocol,
-                          AF_IP(laddr), lp->lp_lport),
-                          SK_PRI_ARGS(other_skf)));
+    if( fm->fm_hwports_no5tuple )
+      rc = -EEXIST;
+    else
+      OO_DEBUG_IPF(other_skf = oof_wild_socket(lp, lpa, skf->af_space);
+                   if( other_skf != NULL )
+                     ci_log(FSK_FMT "STEAL "IPX_TRIPLE_FMT" from "SK_FMT,
+                            FSK_PRI_ARGS(skf),
+                            IPX_TRIPLE_ARGS(lp->lp_protocol,
+                            AF_IP(laddr), lp->lp_lport),
+                            SK_PRI_ARGS(other_skf)));
   }
   if( nat_preimage.n_results > 0 )
     oof_nat_table_lookup_free(&nat_preimage);
--- a/src/lib/efthrm/oof_filters.c
+++ b/src/lib/efthrm/oof_filters.c
@@ -2555,16 +2555,18 @@ oof_socket_add_wild(struct oof_manager* fm, struct oof_socket* skf,
      * stack, resulting in unaccelerated sockets in this stack. Return an error
      * here to prevent this.
      */
-    if( fm->fm_hwports_no5tuple )
+    if( fm->fm_hwports_no5tuple ) {
       rc = -EEXIST;
-    else
+    }
+    else {
       OO_DEBUG_IPF(other_skf = oof_wild_socket(lp, lpa, skf->af_space);
-                   if( other_skf != NULL )
-                     ci_log(FSK_FMT "STEAL "IPX_TRIPLE_FMT" from "SK_FMT,
-                            FSK_PRI_ARGS(skf),
-                            IPX_TRIPLE_ARGS(lp->lp_protocol,
-                            AF_IP(laddr), lp->lp_lport),
-                            SK_PRI_ARGS(other_skf)));
+      if( other_skf != NULL )
+        ci_log(FSK_FMT "STEAL "IPX_TRIPLE_FMT" from "SK_FMT,
+                   FSK_PRI_ARGS(skf),
+                   IPX_TRIPLE_ARGS(lp->lp_protocol,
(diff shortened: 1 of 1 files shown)


sianj-xilinx:
The way that sort of scenario is expected to be handled is through use of the existing onload module parameters `oof_shared_keep_thresh` and `oof_shared_steal_thresh`. In this case I would expect that by setting `oof_shared_steal_thresh=0` you should allow the second stack to steal the existing wild filter. Have you tried this?

mina:
Thanks for the suggestion, I've tried `oof_shared_steal_thresh=0`, and we've root caused a related bug in oof_local_port_addr_fixup_wild that was making the stealing not work on our platform, but long story short, `oof_shared_steal_thresh=0` and our related fix doesn't seem to fully resolve the issue.

The behavior with `oof_shared_steal_thresh=0` and our fix to oof_local_port_addr_fixup_wild, what I see is:

1. One onload stack starts, exits, and leaves behind a hw filter.
2. The next onload stack starts, and oof_socket_add_wild sees that there is an existing hw filter.
3. Asynchronously oof_local_port_addr_fixup_wild will run (I guess it runs when the previous onload stack starts the delayed cleanup). oof_local_port_addr_fixup_wild will delete the hw filter pointing to the old stack, and insert the hw filter to the new stack.

This all works as expected, except in the case where the new onload stack receives data between steps 2 and 3. In this case, I still see some unaccelerated sockets:

```
Every 1.0s: onload_stackdump stats | grep -i accept                                                                                                                                                  almasrymina-3: Mon Jun  2 22:56:49 2025

tcp_accept_os: 573
ul_accepts: 428
```

What happened in this run is that 573 sockets were opened before  oof_local_port_addr_fixup_wild ran, and those were accelerated, and then 428 sockets were open after oof_local_port_addr_fixup_wild and fixed up the hw filters, so those became accelerated.

The solution in this patch still seems valid to me, let me know if you disagree. Basically we make it so that oof_socket_add_wild sees an existing filter pointing to another onload stack, then it will fail creating the new stack, rather than relying to a future call to oof_local_port_addr_fixup_wild to fix the hw filters (and there is a time window where the hw filter is not yet stolen, which means unaccelerated sockets).

We will share the fix we found for the stealing in oof_local_port_addr_fixup_wild to work properly separately.

sianj-xilinx:
I'm sorry I've made you wait so long for this. The original issue that I hoped to bundle this with turned out to be more separate than I'd imagined, so I didn't manage to look at it at that point. I'm now dedicating some time looking explicitly at this issue. I need to give it a bit of thought to ensure that the changes (this and #282) don't break assumptions for non-AF_XDP NICs. I also want to think a bit about implications for our X3 (efct) NICs, which like AF_XDP don't support filter redirect. One option if I can't satisfy myself immediately that things are ok might be for me to add some unit tests verifying this behaviour and to restrict the changes to AF_XDP, so that I can unblock you, but still allow me a bit more time to look at the other hardware, with the unit tests to check I don't re-break AF_XDP when making the later changes. I'll give you an update on what I find tomorrow.

mina:
No worries at all! Thank you for your review and please do try to follow up here. If there is another approach that works, we'd be happy to follow up.

sianj-xilinx:
Looking into this I've been able to trigger corruption of our filter management state with some of the filter steal type scenarios when using AF_XDP. The exact failure mode varies before and after your fix, so I don't think you've made things worse here, but I think there's more to do for a complete fix. What I'd like to do is to overhaul the handling of filter redirection for NICs that don't support it. both AF_XDP and X3. As I'm sure you've noticed, this filtering code is not the simplest or cleanest part of the onload codebase, so this isn't a trivial job.

Given that our AF_XDP support is officially alpha quality anyway, and this improves behaviour for a real use case I'm ok with merging your change here in the mean time while I'm working on a more complete fix.

src/lib/efthrm/oof_filters.c
```diff
@@ -2548,14 +2548,23 @@ oof_socket_add_wild(struct oof_manager* fm, struct oof_socket* skf,
...
+    if( fm->fm_hwports_no5tuple )
+      rc = -EEXIST;
+    else
+      OO_DEBUG_IPF(other_skf = oof_wild_socket(lp, lpa, skf->af_space);
+                   if( other_skf != NULL )
+                     ci_log(FSK_FMT "STEAL "IPX_TRIPLE_FMT" from "SK_FMT,
+                            FSK_PRI_ARGS(skf),
+                            IPX_TRIPLE_ARGS(lp->lp_protocol,
+                            AF_IP(laddr), lp->lp_lport),
+                            SK_PRI_ARGS(other_skf)));
```
sianj-xilinx:
```suggestion
    if( fm->fm_hwports_no5tuple ) {
      rc = -EEXIST;
    }
    else {
      OO_DEBUG_IPF(other_skf = oof_wild_socket(lp, lpa, skf->af_space);
      if( other_skf != NULL )
        ci_log(FSK_FMT "STEAL "IPX_TRIPLE_FMT" from "SK_FMT,
                   FSK_PRI_ARGS(skf),
                   IPX_TRIPLE_ARGS(lp->lp_protocol,
                   AF_IP(laddr), lp->lp_lport),
                   SK_PRI_ARGS(other_skf)));
    }
```

mina:
> Given that our AF_XDP support is officially alpha quality anyway, and this improves behaviour for a real use case I'm ok with merging your change here in the mean time while I'm working on a more complete fix.

Ack, makes sense to me. I think we leave steal at the default configuration (maybe that's disabled) so we haven't reproduced the issue you're referring to. Merging the fix for the reproducible issue sounds good to me. If coures if we run into more issues on our side we'll try to send fixes and maybe we'll work together to mold into the more complete fix. We're happy to help on our side.

mina:
I applied your suggestion, let me know if anything else is needed!

PR #281, opened 2025-05-30
Improve select iterate speed

liujinhui-job:
Traverse a bitmap in an alternative way
It performs particularly well in scenarios where file descriptors are sparsely distributed.

The larger the FD value, the better the performance.
such as:
(if fd is 1000)
fd_set readfds;
FD_ZERO(&readfds);
FD_SET(1000, &readfds);
select(1000 + 1, &readfds, NULL, NULL, NULL);

select achieves a performance boost of over 20x when handling large file descriptor.

### DIFF
--- a/src/lib/transport/unix/poll_select.c
+++ b/src/lib/transport/unix/poll_select.c
@@ -155,7 +155,7 @@ static inline int do_sys_select(const char* why, int nfds,
  * get next fd bit.
  * /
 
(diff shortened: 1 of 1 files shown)


liujinhui-job:
The larger the FD value, the better the performance.
such as:
(if fd is 1000)
fd_set readfds;
FD_ZERO(&readfds);
FD_SET(1000, &readfds);
select(1000 + 1, &readfds, NULL, NULL, NULL);

PR #282, opened 2025-06-02
tcp_filters: Remove filter when efrm_filter_redirect fails

zhuyifei1999:
When efrm_filter_redirect fails with ENOENT or ENODEV, the code path would reach efrm_filter_insert, which would try to insert a filter rule with the same filter but different flow steering target queue.

In the case of AF_XDP, efrm_filter_redirect is not implemented (always ENODEV), and some netdevs return an error for af_xdp_filter_insert (via set_rxnfc) when it detects duplicate flow steering rules, even if the target queue is different, requiring the caller to remove the prior filter first. This patch addresses this.

Without the patch, if filter_insert errors, oofilter->filter_id[hwport] is left as the error code from efrm_filter_redirect and the old rule is forgotten by onload and left dangling.


mina:
Thank you very much. Glad to see we're seeing the same issue. If it's at all possible to give us a preview of the changes you have in mind via a github branch or something, please do. We can help you with the testing.

PR #283, opened 2025-06-03
Do not handle wakeup events if the event queue is not primed

samikhawaja:
Onload stack in kernel for AF_XDP unnecessarlyy tries to handle events even when the event queue is not primed. This wastes cpu cycles and also introduces lock contention when we are running in spinning mode.

Return early and do not handle wakeup events if the event qeueue was not primed when using AF_XDP.

Latency before this change:
latency_p50=0.000015350
latency_p90=0.000021750
latency_p99=0.000023030
latency_p99.9=0.000024630

Latency after this change:
latency_p50=0.000014230
latency_p90=0.000014550
latency_p99=0.000015990
latency_p99.9=0.000017430

### DIFF
--- a/src/lib/efthrm/tcp_helper_resource.c
+++ b/src/lib/efthrm/tcp_helper_resource.c
@@ -6444,13 +6444,22 @@ static int tcp_helper_wakeup(tcp_helper_resource_t* trs, int intf_i, int budget)
 {
   ci_netif* ni = &trs->netif;
   int n = 0, prime_async;
+  struct efhw_nic *nic;
+  bool primed;
 
   TCP_HELPER_RESOURCE_ASSERT_VALID(trs, -1);
   OO_DEBUG_RES(ci_log(FN_FMT, FN_PRI_ARGS(ni)));
   CITP_STATS_NETIF_INC(ni, interrupts);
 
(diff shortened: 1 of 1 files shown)


samikhawaja:
Hi,

I hope you are well.

Please take a look at this when you get time and please let me know if you have any questions or feedback regarding this.

Thanks

sianj-xilinx:
Ideally I'd like to keep architecture specific handling out of the common code. It seems that the problem here arises because onload's idea of primedness differs from AF_XDP's idea, because our AF_XDP implementation effectively always stays primed. An alternative approach might be to address the underlying issue, rather than try and patch it up in the higher layers. To do that would probably mean implementing `af_xdp_nic_wakeup_request()` to add the wait queue only when a wakeup is actually requested, and remove it again in the `wait_callback()`. That should mean that when the app is spinning we're never primed and so don't get the spurious wakeups in the first place.

sianj-xilinx (approved):

samikhawaja:
We found that the event_waiter->evq is always zero for AF_XDP. This should be equal to the NIC queue being used as that is used to do wake ups.

So I added an extra commit in this pull request that fixes this.

Please have a look at that also when you get time.

Thanks

Issue #284, opened 2025-06-06
el10 rpm build issues


lukester1975:
Hello

I've tripped over a couple of issues building for el10/Fedora when doing something like `./scripts/onload_mkpackage -R --version ${VER} --resultdir build/${VER} --user --akmod`.

1. lto is now on by default with rpmbuild, but compilation fails:
```
/usr/include/bits/fcntl2.h:41:1: error: inlining failed in call to ‘always_inline’ ‘open’: function body can be overwritten at link time
   41 | open (const char *__path, int __oflag, ...)
      | ^
/usr/include/bits/fcntl2.h:53:14: note: called from here
   53 |       return __open_alias (__path, __oflag, __va_arg_pack ());
      |              ^
make[4]: *** [/tmp/cc86pX8x.mk:2: /tmp/ccrlprvS.ltrans0.ltrans.o] Error 1
make[4]: *** Waiting for unfinished jobs....
lto-wrapper: fatal error: make returned 2 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed
```
Can be avoided with `%define _lto_cflags %{nil}` in the spec file since it wasn't immediately obvious what the issue was.

2. There is no `.egg-info` created with python>=3.12, just the .egg archive, unless passing `--single-version-externally-managed`:
```
RPM build errors:
    File not found: /path/to/packages/onload-9.0.2-1.el10.src_rpmbuild_rebuild/BUILDROOT/onload-9.0.2-1.el10.x86_64/usr/lib/python3.12/site-packages/sfc*.py
    File not found: /path/to/build/packages/onload-9.0.2-1.el10.src_rpmbuild_rebuild/BUILDROOT/onload-9.0.2-1.el10.x86_64/usr/lib/python3.12/site-packages/__pycache__/sfc*.pyc
    File not found: /path/to/build/packages/onload-9.0.2-1.el10.src_rpmbuild_rebuild/BUILDROOT/onload-9.0.2-1.el10.x86_64/usr/lib/python3.12/site-packages/*Onload*.egg-info
```

My mods are here https://github.com/Xilinx-CNS/onload/compare/v9_0...lukester1975:onload:fix-el10. Happy to make a PR if useful!

Thanks

lukester1975:
Ah the reason I put the `--single-version-externally-managed` check in was:
```
> python3.11 ./scripts/onload_misc/setup.py install --single-version-externally-managed

[...]

PR #287, opened 2025-06-09
Some el10 fixes for #284

lukester1975:


ivatet-amd:
Hi @lukester1975, thank you for all your extremely useful contributions over the past few days!

While we're verifying your Python patch, would you be happy to reconsider the LTO patch from this PR?

AFAIU, the LTO error (`error: inlining failed in call to ‘always_inline’ ‘open’: function body can be overwritten at link time`) is legitimate because `startup.c`, that triggers the LTO error, belongs to the library where Onload intercepts syscalls, and genuinely overrides the `open()`, so that occurrence cannot be hardneded with the "fortify" feature. Instead, in `startup.c` we should use the unambiguous `ci_sys_open()`, also to avoid problems outlined in `citp_basic_syscall_init()`.

Internally, we have a PR with two patches:
```diff
$ git diff HEAD~2
diff --git a/src/lib/transport/unix/startup.c b/src/lib/transport/unix/startup.c
index 7a12c4cf8d..7bcf94dfa9 100644
--- a/src/lib/transport/unix/startup.c
+++ b/src/lib/transport/unix/startup.c
@@ -213,7 +213,7 @@ static void citp_log_to_file(const char *s)
 {
   int fd;
   ci_assert(!CITP_OPTS.log_via_ioctl);
-  fd = open(s, O_WRONLY | O_CREAT | O_TRUNC, S_IREAD | S_IWRITE);
+  fd = ci_sys_open(s, O_WRONLY | O_CREAT | O_TRUNC, S_IREAD | S_IWRITE);
   if( fd >= 0 ) {
     if( citp.log_fd >= 0 )
       ci_sys_close(citp.log_fd);
diff --git a/src/tests/onload/cplane_sysunit/cplane_lib.c b/src/tests/onload/cplane_sysunit/cplane_lib.c
index a352e7b350..92289dfe47 100644
--- a/src/tests/onload/cplane_sysunit/cplane_lib.c
+++ b/src/tests/onload/cplane_sysunit/cplane_lib.c
@@ -4,7 +4,7 @@
 #include <cplane/cplane.h>
 #include <cplane/server.h>

-extern int cplane_ioctl(int, long unsigned int, void* );
+extern int cplane_ioctl(int, long unsigned int, ...);

 int (* ci_sys_ioctl)(int, long unsigned int, ...) = (void*) cplane_ioctl;
```

Would you be able to give it (enabled-by-default-LTO + two patches above) a quick spin to see if it addresses the LTO problem in your setup (and `EF_LOG_FILE` still works OK)? If we keep LTO enabled, it might help us avoid/find more defects in the future.

lukester1975:
Sure, links fine with those patches and log file is generated.

Thanks

ivatet-amd (approved):

abower-amd:
@lukester1975 the upshot of a change just made is that EL's python packaging macros are now used to invoke `setup.py` instead of this invocation from `onload_install`. I would expect this to Do The Right Thing for RHEL but in case it has effectively reversed your change I just thought I would let you know so it can be revisited!

PR #291, opened 2025-06-23
FIX ci_tcp_sendmsg_fill_pkt NULL pointer error

liujinhui-job:
oo_pkt_fill call oo_pkt_filter_next_pkt and ci_netif_pkt_alloc_block，then *p_netif_locked will coredump, bacause it is NULL

### DIFF
--- a/src/lib/transport/ip/tcp_send.c
+++ b/src/lib/transport/ip/tcp_send.c
@@ -171,7 +171,6 @@ int ci_tcp_sendmsg_fill_pkt(ci_netif* ni, ci_tcp_state* ts,
   /* Initialise and fill a packet buffer from an iovec. */
   int n;
   ci_ip_pkt_fmt* pkt = oo_pkt_filler_next_pkt(ni, &sinf->pf, sinf->stack_locked);
-  int p_netif_locked = ci_netif_is_locked(ni);
 
   ci_assert(pkt);
   ci_assert(! ci_iovec_ptr_is_empty_proper(piov));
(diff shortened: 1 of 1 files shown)


liujinhui-job:
While examining the source code, I observed that the parameter "NULL/*p_netif_locked*/" appeared unusual. 

Consequently, I modified the ts->eff_mss value in the ci_tcp_send_fill_pkts function to 4096 and transmitted a 4096-byte packet, which successfully triggered a segmentation fault. 

However, in actual operational scenarios, I'm uncertain whether ts->eff_mss would realistically reach 4096.

ivatet-amd:
I wonder if we are looking at dead code in `oo_pkt_fill()` and should replace `if( pf->buf_start == pf->buf_end )` with `assert( pf->buf_start != pf->buf_end )`. I cannot say now without further looking at the code and various NIC constraints.

src/lib/transport/ip/tcp_send.c
```diff
@@ -171,6 +171,7 @@ int ci_tcp_sendmsg_fill_pkt(ci_netif* ni, ci_tcp_state* ts,
   /* Initialise and fill a packet buffer from an iovec. */
   int n;
   ci_ip_pkt_fmt* pkt = oo_pkt_filler_next_pkt(ni, &sinf->pf, sinf->stack_locked);
+  int p_netif_locked = ci_netif_is_locked(ni);
```
sianj-xilinx, line 174: I haven't looked at the rest of this to check the fix as a whole, but this line is definitely wrong. What you are checking here is whether _anyone_ holds the netif lock. What you want to know is whether the _caller_ holds the netif lock. That information is in `sinf->stack_locked`.

PR #295, opened 2025-06-30
Ensure fdtable seq is assigned atomically

liujinhui-job:
When two threads concurrently invoke the socket function, the absence of
lock protection in citp_fdinfo_init may occasionally result in identical
sequence numbers (seq) being assigned to different fdi entries. 

This change replaces the non-atomic increment with __sync_fetch_and_add
to guarantee uniqueness of seq in a multi-threaded environment.


ol-alexandra (changes requested):
The change itself looks reasonable, but the commit log should include more details, I think.  The description of the pull request includes all these details, but the git commit log does not.

PR #300, opened 2025-07-19
Alternative to legacy distutils

castramet:
distutils was deprecated in Python 3.10 and fully removed in Python 3.12

Systems like rhel 10 install Python 3.12

This change replaces distutils with setuptools


abower-amd:
Thank you for this contribution @castramet! Sorry for the late review. We have now already included this change as part of https://github.com/Xilinx-CNS/onload/commit/f68da3a1ed8b064545b74fc5c388943ce750e4da#diff-3597a9ff339cccfce64639162a851ffc04e47b7cb16d8198e45085ff2f75fc1b, but it would have been better as a separate commit like your change rather than tucked into a Debian-based commit. Sorry about that.

PR #314, opened 2025-12-29
af_xdp: Suffix the XSP map pin paths

zhuyifei1999:
Each XSK and XDP prog are bound to a specific nic, so load the prog multiple times, once for each NIC, and distinguish the XSKMAPs with the ifname.

Updated API from bpf_object__pin_maps to bpf_map__pin in order to specify the exact map to pin and the target file path; the former API would use a filename matching the map name and would not suffix the ifname.

### DIFF
--- a/src/lib/efhw/af_xdp.c
+++ b/src/lib/efhw/af_xdp.c
@@ -32,6 +32,7 @@
 
 #define XDP_PROG_NAME "xdpsock"
 #define BPF_FS_PATH "/sys/fs/bpf/"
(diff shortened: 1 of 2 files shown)


zhuyifei1999:
Oops, this too. Fixed

mina:
Thank you SO MUCH Sian! Very happy to work with you on these contributions! 

If you see any other issues with the other pull requsets as well please let us know and we'll be happy to iterate on them. Thanks!

PR #315, opened 2026-01-14
The onload-dkms package requires libcap-dev.

lukester1975:
Hi

E.g. if installed in a minimal system that didn't generate the debs in the first place, the module build will fail.

I guess there's an argument to include more of the `Build-Depends` if requiring `dkms` (and `build-essential` transiently) doesn't bring in all the required deps, but I've not personally had a problem other than `libcap-dev`.

RPM looks OK due to `user_build_requires` for the akmod spec, AFAICT.

Thanks!


lukester1975:
Sure, makes sense. It's months since I tripped over it so don't recall anything further, but from a quick look I guess `ci_is_sysadmin()`?

Thanks

PR #324, opened 2026-04-16
ON-17300: do not put object file in source dir

okt-kosovnn:
This is a fixup for 99d863f2a08091050438af9e8cdf94fefb133cb5

Without this commit even after deleting a build, the file /src/tools/ip/libstack.o remains present. This is a unique object file that is not located in the /build/ directory.
If we build Onload on one host, remove build and then build on another with different system it may lead to build fail.


sianj-xilinx (approved):
(for onload-9.2)

PR #327, opened 2026-05-18
af_xdp: Perform netdev lock, presence check, ->begin around ethtool ops

zhuyifei1999:
It has been observed that at least one driver would panic when the ethtool ops is performed on a detached netdev [1]. Kernel upstream said that if a module calls the functions "without holding the appropriate locks, then no amount of changes to the [...] driver will fix that issue", and "the root cause that something is calling the drivers ethtool function without correctly holding the expected locks."

This patch duplicates the locking, presence check, and ethtool_ops->begin logic from __dev_ethtool. The pm_runtime_get_sync logic is not copied because we don't expect a onload-attached device to be runtime-suspended; and if it does somehow happen I think returning error is okay.

On kernels without netdev_lock_ops, i.e. prior to this patch series [2], consider RTNL sufficient.

[1] https://patchwork.ozlabs.org/project/intel-wired-lan/patch/20260421051641.370436-1-boolli@google.com/
[2] https://lore.kernel.org/all/20250305163732.2766420-1-sdf@fomichev.me/

### DIFF
--- a/src/lib/efhw/af_xdp.c
+++ b/src/lib/efhw/af_xdp.c
@@ -880,6 +880,8 @@ static int af_xdp_dmaq_kick(struct efhw_nic *nic, int instance)
  * Initialisation and configuration discovery
  *
  *---------------------------------------------------------------------------*/
+static int af_xdp_pre_ethtool(struct efhw_nic *nic);
+static void af_xdp_post_ethtool(struct efhw_nic *nic);
 /* Update the efhw_nic struct with the nic's supported RSS hash key length
  * and indirection table length. */
 static int
@@ -891,12 +893,16 @@ af_xdp_rss_get_support(struct efhw_nic *nic)
 
 	ASSERT_RTNL();
 
+	rc = af_xdp_pre_ethtool(nic);
+	if (rc)
+		return rc;
+
 	ops = dev->ethtool_ops;
 	if (!ops->get_rxfh_indir_size) {
 		EFHW_WARN("%s: %s does not support `get_rxfh_indir_size` operation",
 							__FUNCTION__, dev->name);
 		rc = -EOPNOTSUPP;
-		goto unlock_out;
+		goto out_ethtool;
 	}
 
 	nic->rss_indir_size = ops->get_rxfh_indir_size(dev);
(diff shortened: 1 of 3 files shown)


src/lib/efhw/af_xdp.c
```diff
@@ -1276,20 +1286,65 @@ static enum efhw_page_map_type af_xdp_buffer_map_type(struct efhw_nic *nic)
...
  *--------------------------------------------------------------------*/
 
+static int af_xdp_pre_ethtool(struct efhw_nic *nic)
+{
+  struct net_device *dev = nic->net_dev;
+  int rc;
+
+#ifdef EFRM_HAVE_NETDEV_LOCK_OPS
+  netdev_lock_ops(dev);
+#endif
```
sianj-xilinx: Rather than using including #ifdeffery here (and the other places), could you instead add appropriate `efrm_netdev_lock_ops()` and `efrm_netdev_unlock_ops()` in `include/ci/driver/kernel_compat.h` so we can keep the compat stuff separate from the functional code? Ideally with a comment on which kernel introduced the requirement.

src/lib/efhw/af_xdp.c
```diff
@@ -1497,9 +1564,17 @@ af_xdp_filter_remove(struct efhw_nic *nic, int filter_id)
 	info.fs.location = filter_id;
 
 	rtnl_lock();
+
+	rc = af_xdp_pre_ethtool(nic);
+	if (WARN_ON(rc))
+		goto out_unlock;
```
sianj-xilinx: `WARN_ON` feels a bit harsh here. I'd be inclined to just log this with `EFHW_ERR`, but if you'd prefer to keep the backtrace I could live with `WARN_ON_ONCE`.

src/lib/efhw/af_xdp.c
zhuyifei1999: Hmm, would it be okay if I include `#include <net/netdev_lock.h>` there?

sianj-xilinx: Should be I think, as long as it's protected by compat guards (you'll need `autocompat.h` first).

zhuyifei1999: Done

src/lib/efhw/af_xdp.c
zhuyifei1999: Done

zhuyifei1999:
Let me test this new version, hold on

Edit: seems to work

sianj-xilinx (approved):

PR #331, opened 2026-06-03
deb: DKMS packaging improvements

changchinlan:
I've been packaging onload as .deb for internal use (Ubuntu 20.04 through 24.04) and ran into a few issues with the current debian templates that this series tries to address.

Problems I hit:

- DKMS modules weren't making it into the initramfs, so in-tree sfc would load instead of the DKMS-built one at boot.
- No way to customise the DKMS build (extra dkms.conf options, quilt patches, Makefile override) without forking the templates.
- The postinst/prerm scripts hand-roll dkms calls instead of using dh_dkms.
- pybuild calls are unconditional, which breaks builds for <=9.1 (no pyproject.toml).

The commits are split roughly by feature:

- `env pwd` portability fix (minor, separate concern)
- Configurable debian revision via `--debian-revision` / `DEBIAN_REVISION`
- Hooks for `debian/Makefile`, `debian/dkms.append.conf`, and  `debian/patches` applied into the DKMS source tree. dkms.append.conf lets downstream extend dkms.conf at build time without patching upstream (e.g. declaring BUILD_DEPENDS on out-of-tree drivers like xilinx-efct, or setting BUILD_EXCLUSIVE_KERNEL).
- dh_dkms integration with backwards compat for 20.04
- initramfs-tools hook + modules-load.d + dpkg trigger on xilinx-efct-dkms
- Guard pybuild behind pyproject.toml existence check. Restore `--python-layout=deb` for older releases that still use setup.py.

Tested with onload 9.1.0.24 on Ubuntu 24.04. All five packages built successfully.

Ubuntu 26.04 support: The initramfs hook targets initramfs-tools (24.04 and earlier). Ubuntu 26.04 defaults to dracut, which handles DKMS module priority natively via updates/dkms/, so a dracut module may not be needed.

### DIFF
--- a/scripts/debian/debian-templ/rules
+++ b/scripts/debian/debian-templ/rules
@@ -188,7 +188,11 @@ ifneq (,$(filter #TYPE#-source,$(DOPACKAGES)))
 	dh_installdirs -p$(psource)  usr/src/modules/$(sname)/debian
 
 	# Copy only the driver source to the proper location
-	cp Makefile  debian/$(psource)/usr/src/modules/$(sname)
+	if [ -f "debian/Makefile" ]; then \
+		cp debian/Makefile  debian/$(psource)/usr/src/modules/$(sname); \
+	else \
+		cp Makefile  debian/$(psource)/usr/src/modules/$(sname); \
+	fi
 	cp -r src  debian/$(psource)/usr/src/modules/$(sname)
 	cp -r scripts  debian/$(psource)/usr/src/modules/$(sname)
 	cp -r mk  debian/$(psource)/usr/src/modules/$(sname)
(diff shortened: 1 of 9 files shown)


scripts/debian/debian-templ/rules
```diff
@@ -333,6 +333,7 @@ ifneq (,$(filter #TYPE#-user,$(DOPACKAGES)))
 
 	i_prefix=$(CURDIR)/debian/$(userpackage) scripts/onload_install --verbose \
 		--packaged --userfiles --modprobe --modulesloadd --udev \
+		$(if $(wildcard pyproject.toml),,--python-layout=deb) \
```
abower-amd, line 336: How will this help if it is for older releases as this code is only for newer releases by definition? Or did we miss  compat for older OSs? Bear in mind the current master branch targets Debian 12+ and Ubuntu 24.04 plus and we'd probably happily bump to Trixie if we found anything that needed it.

scripts/mmaketool
```diff
@@ -215,7 +215,7 @@ kernelid() {
 }
 
 do-find-top() {
-  mytoppath=$(cd "$bin/.." && /bin/pwd)
+  mytoppath=$(cd "$bin/.." && /usr/bin/env pwd)
```
abower-amd: For compatibility with what? `pwd` is a POSIX API command defined to be at `/bin/pwd`, regardless of what the OS might do in respect to _usrmerge_. Is there a particular scenario you have in mind where the original path would fail? Otherwise I see no need to invoke an additional tool.

changchinlan: Dropped. That was a habit from working on NixOS where /bin/pwd doesn't exist, but it's POSIX and onload doesn't need to worry about that.

scripts/debian/debian-templ/rules
changchinlan: The motivation is to reuse the same debian templates across multiple onload versions without maintaining a separate fork of the templates (as we are still packaging 8.1.3.40 and 9.1.0.24 dkms alongside master). I'm happy to drop them if that's not something upstream wants to carry.

PR #335, opened 2026-07-20
Implement "explicit wild" endpoints

zhuyifei1999:
It's been observed that for some NICs, flow steering configuration (which is needed for AF_XDP) is a slow control-path process, taking milliseconds. And as a result, opening a listening port and closing causes unreasonable delays in the processing, especially with the `ci_netif_lock(ep->netif)` lock held for the entire duration. For outgoing sockets, there exist the workaround of "active wilds" (EF_TCP_SHARED_LOCAL_PORTS), but for listeners, no such workaround exist.

This patch adds a new state CI_TCP_STATE_EXPLICIT_WILD = CI_TCP_STATE_ACTIVE_WILD | CI_TCP_STATE_NOT_CONNECTED, turning CI_TCP_STATE_ACTIVE_WILD sort of into a bitmasked value. Common code between active wilds and explicit wilds now check for that value after masking.

The new endpoint state is always orphaned (no fd), only creatable on stack creation, and is always non-matching for packet delivery to endpoints (allowing the packet to be sent to kernel sockets listening on the same port). It also never has a kernel socket attached, this is so that someone can connect to it and see that the the connection is refused, rather than being delivered to a socket that will not respond. As a result, user has to explicitly make sure that those sockets are not used for other outgoing sockets via sysctls.

User passes the list of ports for explicit wilds in the env variable `EF_TCP_EXPLICIT_WILD_PORTS`, in the format of what bitmap_parselist accepts.

### DIFF
--- a/src/include/ci/internal/ip.h
+++ b/src/include/ci/internal/ip.h
@@ -1180,9 +1180,13 @@ extern int ci_pipe_list_to_iovec(ci_netif* ni, struct oo_pipe* p,
 /* Set in a socket that is used as the owner for an active wild filter */
 #define CI_TCP_STATE_ACTIVE_WILD (0xe000)
 
+/* Set in a socket that is used as the owner for an explicit wild filter */
+#define CI_TCP_STATE_EXPLICIT_WILD (0xe000 | CI_TCP_STATE_NOT_CONNECTED)
+
 
 /* Convert state to number in range 0->0xe */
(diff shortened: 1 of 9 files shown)


zhuyifei1999:
I can't think of a better name for this than "explicit wild", so feel free to suggest other names. Also not sure if CI_TCP_STATE_ACTIVE_WILD into a bitmask is the right choice. I didn't want to use up the last type slot.

PR #336, opened 2026-07-20
Limit filter hwports mask to only interfaces with the filtered IP

zhuyifei1999:
Ths change adds a module option to, when a filter is installed, the filter is only installed to interfaces with the IP of that filter. For wild sockets, Onload duplicates the filter for each local address. The original behavior would be that each such filter would be installed to every active interface. Assuming N interfaces and each interface has a different IP, this would result in N filters being installed to each interface, and N^2 filters installed in total. With this option on, however, each interface would get only one filter and N filters being installed in total.

When new local addresses are being added, the code will retroactively add apply it existing filters, just like original behavior. However, if an existing local address of one interface is added to a new interface, the original code ignores it, since it assumes that the filter for that existing local address has already been added to all active interfaces. This assumption no longer holds with this patch, so the filters will be re-applied.

For local address removal, original code postpones filter removal until the local address is unused on all active interfaces. This behavior is preserved because local address removal is a rare operation not worth optimizing for, in my opinion.

The module option is defaulted off for backwards compatibility, in case a user depends on the original behavior, such as relying on a connection being accelerated even when it is received on an interface when the destination address corresponds to a different interface.

Multicast logic is untouched. DNAT logic I was unable to test.

### DIFF
--- a/src/driver/linux_onload/driver.c
+++ b/src/driver/linux_onload/driver.c
@@ -120,6 +120,13 @@ MODULE_PARM_DESC(oof_use_all_local_ip_addresses,
                  "of the network interface type.\n"
                  "See also cplane_use_prefsrc_as_local module option.");
 
+module_param(oof_limit_filter_to_interface_ip, int, S_IRUGO | S_IWUSR);
+MODULE_PARM_DESC(oof_limit_filter_to_interface_ip,
+                 "By default Onload installs filters for each local IP address "
+                 "to every Onloadable network interface. This option limits "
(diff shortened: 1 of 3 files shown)


zhuyifei1999:
This patch complements PR #335 so fewer filters are created in multi-NIC machines where those NICs have different IPs

mina:
Hi folks!

This is a somewhat critical issue & fix for users. If you could take a look that would be much appreciated! Thanks! We're available to address any issues you may find.

